PR #1, opened 2022-12-31
Split jsons per pack so we don't end up with a super json nightmare

erasels:


modargo:
The key change here is that now each pack must have its own json files in a folder named the pack class but all lowercase (so DownfallPack goes in the folder downfallpack). The only exception is for card packs of entirely base game cards, which are listed in a central place and skipped over. This is all implemented by piggy-backing off the very helpful reflection logic in AutoAdd.

Alongside that, packs are now automatically registered -- once I hooked up AutoAdd to load per-pack JSON, hooking it up to register the packs was very easy. I'll add a few comments in specific areas of interest.

src/main/java/thePackmaster/SpireAnniversary5Mod.java
```diff
@@ -162,50 +164,87 @@ public void receiveEditCards() {
...
+        for (CtClass packClass : packClasses) {
+            String packName = packClass.getSimpleName().toLowerCase();
+            String languageAndPack = getLangString() + "/" + packName;
+            BaseMod.logger.info("Loading pack strings for pack " + packClass.getName() + ". Strings expected to be in folder Resources/localization/" + languageAndPack);
+            BaseMod.loadCustomStringsFile(CardStrings.class, modID + "Resources/localization/" + languageAndPack + "/Cardstrings.json");
+            BaseMod.loadCustomStringsFile(RelicStrings.class, modID + "Resources/localization/" + languageAndPack + "/Relicstrings.json");
+            BaseMod.loadCustomStringsFile(PowerStrings.class, modID + "Resources/localization/" + languageAndPack + "/Powerstrings.json");
+            BaseMod.loadCustomStringsFile(UIStrings.class, modID + "Resources/localization/" + languageAndPack + "/UIstrings.json");
+            BaseMod.loadCustomStringsFile(StanceStrings.class, modID + "Resources/localization/" + languageAndPack + "/Stancestrings.json");
+            BaseMod.loadCustomStringsFile(OrbStrings.class, modID + "Resources/localization/" + languageAndPack + "/Orbstrings.json");
```
modargo, line 205: I decided that it was simplest to require that each pack have all six of these files. It's very easy to copy over empty ones and just ignore the ones a particular mod doesn't use.

src/main/java/thePackmaster/SpireAnniversary5Mod.java
```diff
@@ -162,50 +164,87 @@ public void receiveEditCards() {
...
-
         BaseMod.loadCustomStringsFile(OrbStrings.class, modID + "Resources/localization/" + getLangString() + "/Orbstrings.json");
+
+        loadPackStrings();
+    }
+
+    // These packs are excluded from loading of pack-specific string files because they consistent entirely of base game cards.
+    // If you're making a pack that also consists of only base game cards, add it to this list.
+    // The name and description of the pack can go in the main UIstrings.json file.
+    private static final List<String> baseGamePacks = Arrays.asList(IroncladPack.class.getName(), SilentPack.class.getName(), DefectPack.class.getName(), WatcherPack.class.getName());
```
modargo, line 186: This is the one central line of code that could result in merge conflicts now, but there are very few packs with only base game cards so I doubt it will be an issue.

src/main/resources/anniv5Resources/localization/eng/Cardstrings.json
```diff
@@ -65,275 +65,5 @@
   "${ModID}:FromMyHat": {
     "NAME": "From My Hat",
     "DESCRIPTION": "Gain !B! Block. NL Add a random Attack into your hand. NL It gains Retain."
-  },
```
modargo, line 68: The remaining entries in this file are for the starter cards and cards for a demo pack that appears to have been commented out. Probably good to remove those entries at some point, but I didn't want to take too much of a detour from the main point of this PR.

src/main/java/thePackmaster/cardmodifiers/madsciencepack/ApplyFrosbiteModifier.java
```diff
@@ -14,7 +15,7 @@ public ApplyFrosbiteModifier(int valueIn){
 
     @Override
     public String modifyDescription(String rawDescription, AbstractCard card) {
-        return rawDescription + CardCrawlGame.languagePack.getUIString("MadScienceModifiers").TEXT[1];
+        return rawDescription + CardCrawlGame.languagePack.getUIString(SpireAnniversary5Mod.makeID("MadScienceModifiers")).TEXT[1];
```
modargo, line 18: All the changes like this are because there was an entry that wasn't prefixed with the mod ID.

src/main/java/thePackmaster/patches/OpeningRunScreenPatch.java
```diff
@@ -15,8 +17,10 @@ public class OpeningRunScreenPatch {
             locator = Locator.class
     )
     public static void SetTheThing(NeowEvent __instance) {
-        BaseMod.logger.info("Packmaster is Ready To Do Thing");
-        SpireAnniversary5Mod.readyToDoThing = true;
+        if (AbstractDungeon.player.chosenClass == ThePackmaster.Enums.THE_PACKMASTER) {
```
modargo, line 20: This was an unrelated issue I fixed along the way -- the packmaster logic was being run for all characters, so if you tried to start a run with a non-packmaster character it would crash.

src/main/java/thePackmaster/SpireAnniversary5Mod.java
```diff
@@ -162,50 +164,87 @@ public void receiveEditCards() {
     @Override
     public void receivePostInitialize() {
         declarePacks();
+        BaseMod.logger.info("Full list of packs: " + allPacks.stream().map(pack -> pack.name).collect(Collectors.toList()));
```
modargo, line 167: I added in a few logs like this that were helpful for my debugging. I've left them in for this PR because I think they may be helpful during the development process -- they may help work out any issues that people contributing to this run into, or at least confirm that their pack is being registered. Feel free to remove later if they're no longer helpful.

src/main/java/thePackmaster/SpireAnniversary5Mod.java
```diff
@@ -162,50 +164,87 @@ public void receiveEditCards() {
...
+            List<com.evacipated.cardcrawl.mod.stslib.Keyword> packKeywords = new ArrayList<>(Arrays.asList(gson.fromJson(packJson, com.evacipated.cardcrawl.mod.stslib.Keyword[].class)));
+            keywords.addAll(packKeywords);
+        }
 
-        if (keywords != null) {
-            for (Keyword keyword : keywords) {
-                BaseMod.addKeyword(modID, keyword.PROPER_NAME, keyword.NAMES, keyword.DESCRIPTION);
-            }
+        for (Keyword keyword : keywords) {
+            BaseMod.addKeyword(modID, keyword.PROPER_NAME, keyword.NAMES, keyword.DESCRIPTION);
```
modargo, line 231: Note that all keywords (and all other strings too) get loaded into the same place, so as long as a string is present in _any_ of the JSON files, using that string will work. In other words, the separation is purely for organizational purposes.

erasels:
Great explanation and contribution, thank you. I think requiring people to create empty files may be the wrong decision here though. I don't know if there will be a compile time impact or not, but I think this can be easily fixed.

I'll make that addition to the system myself though.

modargo:
> Great explanation and contribution, thank you. I think requiring people to create empty files may be the wrong decision here though. I don't know if there will be a compile time impact or not, but I think this can be easily fixed.
> 
> I'll make that addition to the system myself though.
Yeah it's easy enough to change that if you want. Just check for the existence of each file and skip any that aren't there. IMO it doesn't make much of a difference; the upside is trivial (anyone making a pack should just copy over another pack's folder), the downside is that issues with misnamed or missing files are less clear.

I'd suggest still requiring a cards JSON file for every pack, because I don't think there's any legitimate use case for a pack not having any cards. Also, that will mean you get an error on load if you name your folder or your cards json file incorrectly, which is better because it makes the problem clear.

PR #14, opened 2023-01-02
Add "Biting Cold" pack

ActuallyEnbeon:
Includes Frostbite (power) & Snowglobe (pack-exclusive relic)


src/main/java/thePackmaster/actions/bitingcoldpack/ActivateGlaciateAction.java
```diff
@@ -0,0 +1,30 @@
...
+import com.megacrit.cardcrawl.actions.AbstractGameAction;
+import com.megacrit.cardcrawl.actions.utility.WaitAction;
+import com.megacrit.cardcrawl.core.Settings;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.powers.AbstractPower;
+import com.megacrit.cardcrawl.powers.watcher.VigorPower;
+
+import static thePackmaster.util.Wiz.*;
+
+public class ActivateGlaciateAction extends AbstractGameAction {
```
erasels, line 12: Why is this an action? Just because you wanted to be absolutely sure the power icon flashes at the correct time? 
You can put that into an anonymous action.

src/main/java/thePackmaster/actions/bitingcoldpack/FrostbiteDamageAction.java
```diff
@@ -0,0 +1,68 @@
...
+    @Override
+    public void update() {
+        if(shouldCancelAction()) {
+            this.isDone = true;
+        } else {
+            tickDuration();
+            if (this.isDone) {
+
+                // Snowglobe
+                if (AbstractDungeon.player.hasRelic(Snowglobe.ID)) {
```
erasels, line 44: Just a small tip, you can store the getRelic return in an AbstractRelic variable and check if it's null. This saves you a list iteration through player.relics

src/main/java/thePackmaster/packs/BitingColdPack.java
```diff
@@ -0,0 +1,26 @@
...
+    public static final String AUTHOR = UI_STRINGS.TEXT[2];
+
+    public BitingColdPack() {
+        super(ID, NAME, DESC, AUTHOR);
+    }
+
+    @Override
+    public ArrayList<String> getCards() {
+        ArrayList<String> cards = new ArrayList<>();
+        // cards.add(card.ID);
```
erasels, line 23: You didn't add your cards to the pack

src/main/java/thePackmaster/patches/bitingcoldpack/OnApplyDebuffPatch.java
```diff
@@ -0,0 +1,44 @@
...
+import javassist.CtBehavior;
+import thePackmaster.actions.bitingcoldpack.ActivateGlaciateAction;
+import thePackmaster.cards.bitingcoldpack.GrowingAffliction;
+import thePackmaster.powers.bitingcoldpack.FrostbitePower;
+import thePackmaster.powers.bitingcoldpack.GlaciatePower;
+
+import java.lang.reflect.Field;
+
+// Patch for effects that activate upon the player applying a debuff
+@SpirePatch(clz = ApplyPowerAction.class, method = "update")
```
erasels, line 19: This could be done in `receivePostPowerApplySubscriber` in SpireAnniversary5Mod
That way we don't reinvent the wheel for the same thing multiple times.

src/main/java/thePackmaster/powers/bitingcoldpack/FrostbitePower.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    public void updateDescription() {
+        description = DESCRIPTIONS[0] + this.amount + DESCRIPTIONS[1];
+    }
+
+    @Override
+    public int getHealthBarAmount() {
+        if(AbstractDungeon.player.hasRelic(SpireAnniversary5Mod.makeID("Snowglobe"))) {
```
erasels, line 62: You're passing a string literal with extra steps here. You could instead use Snowglobe.ID, this way you can make sure that the ID will be the same even if it's changed later on.

erasels (changes requested):
Great stuff, just pointed out a few things that could be improved upon

ActuallyEnbeon:
Ok, I think I've made all the suggested changes. Hopefully should be good to go now!

PR #18, opened 2023-01-02
Shaman card pack

modargo:
The first of my packs.

It's all pretty straightforward and should be within all the guidelines. I'll add comments for a few things that are noteworthy.

pack cards


src/main/java/thePackmaster/cards/AbstractPackmasterCard.java
```diff
@@ -82,18 +82,7 @@ protected Texture getPortraitImage() {
...
-            case ATTACK:
-            case POWER:
-            case SKILL:
-                textureString = makeImagePath("cards/" + cardName + ".png");
-                break;
-            default:
-                textureString = makeImagePath("ui/missing.png");
-                break;
-        }
+        String textureString = makeImagePath("cards/" + cardName + ".png");
```
modargo, line 85: This is the fix I mentioned to you. There's no reason to have a switch on cardtype here at all.

src/main/java/thePackmaster/patches/shamanpack/FineTuneLineWidthPatch.java
```diff
@@ -0,0 +1,53 @@
...
+import javassist.expr.NewExpr;
+import thePackmaster.cards.shamanpack.FreezeAndBurn;
+import thePackmaster.cards.shamanpack.FueledByEmbers;
+
+@SpirePatch(clz = AbstractCard.class, method = "initializeDescription")
+public class FineTuneLineWidthPatch {
+    public static final float WIDTH_ADJUSTMENT = 20.0f;
+    public static boolean addWidthBack = false;
+
+    private static boolean matches(AbstractCard card, String word) {
```
modargo, line 18: Why does this patch exist, you ask? Here is exhibit A.

Without the patch, absolutely horrific line breaks:
card without patch

With the patch, bliss:
card with patch

Note that this is restricted to handle only specific cards, so there's no risk of it affecting other cards or other packs. I've done something similar in my other mods without any issues.

modargo: Oh, and the other card that needing fine-tuning was Freeze and Burn, which had "enemies" on its own line as the final line of the card. The classic poor line break for "Deal N damage to ALL enemies."

modargo:
Okay, that should be all the comments. You should be able to merge this once you've looked it over.

PR #19, opened 2023-01-02
Add "Into the Breach" Pack

ActuallyEnbeon:
Here's my second pack! Should all be ok but do say if there are any problems.
(This PR also includes a couple of updates to my code comments from the Biting Cold pack - don't mind them, they're really just for the benefit of other people looking at my code)


src/main/java/thePackmaster/cards/intothebreachpack/ReboundVolley.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public void branchUpgrade() {
+        upgradeDamage(-2);
+        upgradeMagicNumber(1);
+        rawDescription = CardCrawlGame.languagePack.getCardStrings(ReboundVolley.ID).EXTENDED_DESCRIPTION[0];
+    }
+
+    @Override
+    public void calculateCardDamage(AbstractMonster mo) {
+        super.applyPowers();
```
erasels, line 51: Usually you call the method you're overwriting at the start, then change variables then call the method again.
I'm not sure if applyPowers is the correct call here

```suggestion
        super.calculateCardDamage(mo);
```

src/main/java/thePackmaster/cards/intothebreachpack/SpartanShield.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        blck();
+        applyToSelf(new StrengthPower(p, magicNumber));
+        applyToSelf(new LoseStrengthPower(p, magicNumber));
+    }
+
+    public void upp() {
+        upgradeBlock(2);
```
erasels, line 28: Not an error but might be cool to upgrade the str gain by 1 as well? Don't have to do that but a 2 block upgrade is kind fo sad on its own.

src/main/java/thePackmaster/powers/intothebreachpack/StormGenPower.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public StormGenPower(AbstractCreature owner, int amount) {
+        super(POWER_ID, NAME, PowerType.BUFF, false, owner, amount);
+    }
+
+    @Override
+    public void atEndOfTurn(boolean isPlayer) {
+        ArrayList<AbstractMonster> withDebuff = new ArrayList<>();
+
+        for (AbstractMonster m : AbstractDungeon.getCurrRoom().monsters.monsters)
```
erasels, line 30: This iterates through dead monsters as well.
```suggestion
        for (AbstractMonster m : Wiz.getEnemies())
```
This requires a new import, be sure to add that

src/main/java/thePackmaster/cards/intothebreachpack/ReboundVolley.java
erasels: Wait, what are you even doing here? secondDamage is a custom variable and already scales like damage, what do you even need to change calculateCardDamage for? Furthermore, only overriding this method and not applyPowers means this change is only applied when the player has dragged the card and is about to play it.

Please explain

erasels (changes requested):

src/main/java/thePackmaster/cards/intothebreachpack/ReboundVolley.java
ActuallyEnbeon: So this might be a bit janky, but the aim here was to make secondDamage only change with relation to the player's powers, relics, etc. and not with the targeted monster's. Since secondDamage is used in the cardtext to show the "random enemy" damage, I felt like it makes more sense for it to not change when a specific enemy is targeted, since the true damage value may end up being different from that, so it provides a more accurate picture of how much damage the card will deal.

src/main/java/thePackmaster/cards/intothebreachpack/SpartanShield.java
ActuallyEnbeon: Oh thanks for catching that, I meant to put it a Strength upgrade but I must have forgotten ^^

src/main/java/thePackmaster/cards/intothebreachpack/ReboundVolley.java
ActuallyEnbeon: Further to my point, a card like Rip and Tear doesn't change based on any enemy's Vulnerable - this happens anyway since it's set to target ALL_ENEMY - so I thought it made sense to match that here, given that the effects are similar.

erasels: Ah, I see where you're coming from. This is an interesting consideration to make.

I hope you've tested this thoroughly because this seems like the type of code that could easily cause some weirdness.

ActuallyEnbeon: I have indeed.
Also, looking at AbstractCard's applyPowers and calculateCardDamage methods, they appear to be functionally identical - aside from the latter method checking the target monster - so there shouldn't be any unexpected interactions.
(That isn't to say that I haven't made _sure_ that there aren't, of course. Always better safe than sorry.)

PR #40, opened 2023-01-05
Infest pack, tokens are colorless, remove some testing cards from the starter deck

DarkVexon:
it's epic! it's gaming! it's... epic gaming!


src/main/java/thePackmaster/powers/infestpack/LordOfTheFliesPower.java
```diff
@@ -0,0 +1,40 @@
...
+    public static final String[] DESCRIPTIONS = powerStrings.DESCRIPTIONS;
+
+    public LordOfTheFliesPower(AbstractCreature owner, int amount) {
+        super(POWER_ID, NAME, PowerType.BUFF, false, owner, amount);
+        AbstractDungeon.player.gameHandSize += amount;
+    }
+
+    @Override
+    public void onRemove() {
+        AbstractDungeon.player.gameHandSize -= amount;
```
erasels, line 27: This will not be called when the player wins combat with this power on them. You'll also need to use onVictory

erasels (changes requested):

PR #42, opened 2023-01-05
Aggression pack and a few general fixes

modargo:
Pack 3 (of maybe 4). A pretty straightforward one.

I've also included three small general changes in this PR:
* Preventing crashes if you load a run without the new save field
* Setting the first dropdown for custom draft to the core set pack (instead of the alphabetically first pack)
* Setting which card shows up in the match and keep event (people are welcome to change this to the other starter card if they want, but it's better to have something so that the Match and Keep event works properly)

aggression pack cards


src/main/java/thePackmaster/ThePackmaster.java
```diff
@@ -130,8 +130,7 @@ public String getLocalizedCharacterName() {
 
     @Override
     public AbstractCard getStartCardForEvent() {
-        System.out.println("YOU NEED TO SET getStartCardForEvent() in your " + getClass().getSimpleName() + " file!");
-        return null;
+        return new Rummage();
```
modargo, line 133: I picked the attack rather than the skill. The base game does the same for every character other than Defect, who doesn't have a starter attack (Bash for Ironclad, Neutralize for Silent, Eruption for Watcher). To be fair, Ironclad doesn't have a starter skill, so the precedent is really just Silent and Watcher.

src/main/java/thePackmaster/patches/MainMenuUIPatch.java
```diff
@@ -84,7 +90,7 @@ public class MainMenuUIPatch {
             dropdowns.add(d);
 
             if (i == 0) {
-                d.setSelectedIndex(2);
+                d.setSelectedIndex(coreSetPackIndex);
```
modargo, line 93: This code must have been written assuming that the core set pack was always the first one in the `SpireAnniversary5Mod.allPacks` list. That was true before automatic pack registration, but isn't any more.

Those changes were being made around the same time (in the prep work ahead of opening the repo up to contributions), so this likely worked locally but was broken as soon as it and automatic pack registration were both on the main branch.

PR #57, opened 2023-01-07
Summon GK

bkerns724:


src/main/java/thePackmaster/relics/summonspack/BlueSkull.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public void atBattleStart() {
+        isActive = false;
+        addToBot(new AbstractGameAction() {
+            public void update() {
+                if (!isActive && adp().isBloodied) {
+                    flash();
+                    pulse = true;
+                    adp().addPower(new FocusPower(adp(), FOCUS_AMOUNT));
```
erasels, line 31: Why do this instead of applyPowerAction?

src/main/java/thePackmaster/relics/summonspack/BlueSkull.java
```diff
@@ -0,0 +1,70 @@
...
+            att(new ApplyPowerAction(adp(), adp(), new FocusPower(adp(), FOCUS_AMOUNT), FOCUS_AMOUNT));
+            att(new RelicAboveCreatureAction(adp(), this));
+            isActive = true;
+            adp().hand.applyPowers();
+        }
+
+    }
+
+    public void onNotBloodied() {
+        if (isActive && AbstractDungeon.getCurrRoom().phase == AbstractRoom.RoomPhase.COMBAT)
```
erasels, line 54:
```suggestion
        if (isActive && Wiz.isInCombat())
```
Current check can crash theoretically

erasels (changes requested):

PR #63, opened 2023-01-07
Prismatic pack

modargo:
Pack 4 of 4. Though there's a collaboration with Diamsword that might also make it in.

Nothing out of the ordinary here, though there's a few classes that could in theory be useful for others (AbstractChooseOneAction for making Discovery-like cards, PrismaticUtil for generating cards of any color with conditions).

Also included in this PR: small tweaks to my previous packs (a bug fix and a balance change).

prismatic pack cards


Darkglade1:
Isn't Sword and Dagger supposed to cost 1? Also maybe nerf it to do 7 damage instead (I know Blade Dance exists and is a thing but at least Blade Dance has the downside of pissing off Gremlin Nob)

PR #66, opened 2023-01-07
Adding pack filtering

Pandemonium14:
As I said, UI needs some work for other resoltions


Pandemonium14:
I think that's all now, it's done

PR #79, opened 2023-01-08
- hermitpack implemented

AlexMdle:


src/main/java/thePackmaster/cards/hermitpack/DeadOrAlive.java
```diff
@@ -0,0 +1,49 @@
...
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        if (this.energyOnUse < EnergyPanel.totalCount) {
+            this.energyOnUse = EnergyPanel.totalCount;
+        }
+
+        int num = energyOnUse;
+
+        if (p.hasRelic("Chemical X")) {
+            num += 2;
+            p.getRelic("Chemical X").flash();
```
erasels, line 34: XCost logic in card's use is a nono

src/main/java/thePackmaster/cards/hermitpack/Headshot.java
```diff
@@ -0,0 +1,67 @@
...
+    private static final int UPGRADE_PLUS_DMG = 2;
+
+    public Headshot() {
+        super(ID, 1, CardType.ATTACK, CardRarity.COMMON, CardTarget.ENEMY);
+        baseDamage = DAMAGE;
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        int dam = this.damage;
```
erasels, line 27: What's happening here?

src/main/java/thePackmaster/cards/hermitpack/Headshot.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Override
+    public void calculateCardDamage(AbstractMonster mo)
+    {
+        super.calculateCardDamage(mo);
+
+        if (isDeadOnPos()) {
+            int base_dam = this.damage;
+            this.damage += base_dam;
```
erasels, line 47: Why not just multiply damage by 2 `damage *= 2;`

src/main/java/thePackmaster/cards/hermitpack/Quickdraw.java
```diff
@@ -0,0 +1,48 @@
...
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        AbstractDungeon.actionManager.addToBottom(
+                new DamageAction(m, new DamageInfo(p, damage, damageTypeForTurn), AbstractGameAction.AttackEffect.BLUNT_LIGHT)
+        );
+        this.addToBot(new DrawCardAction(magicNumber));
+
+        DrawReductionPower pow = new DrawReductionPower(p,DRAWLESS);
+        ReflectionHacks.setPrivate(pow,DrawReductionPower.class,"justApplied",false);
+
+        this.addToBot(new ApplyPowerAction(p, p, pow, DRAWLESS));
```
erasels, line 40: When justApplied is false, this gets reduced atEndOfRound, DRAWLESS is 1, so this gets removed at the end of the round before you even draw your cards for the next round and does nothing?

erasels (changes requested):

src/main/java/thePackmaster/cards/hermitpack/Headshot.java
AlexMdle: Remnant from when Snipe was a factor, shouldn't matter either way now, but will shorten.

src/main/java/thePackmaster/cards/hermitpack/Quickdraw.java
AlexMdle: I don't remember the particulars, but it has to do with the strange way draw reduction debuff works. Without this, it does not get reduced at the start of your turn until a turn later.

src/main/java/thePackmaster/cards/hermitpack/Headshot.java
AlexMdle: No idea.

erasels (approved):

PR #91, opened 2023-01-09
Curses/Bells pack

DarkVexon:
Full code and art for the new Bellord pack, which features 10 cards and a keyword Pickup. Some cards give Curses of the Bell on pickup in exchange for high power, while others capitalize on cursed cards.


src/main/java/thePackmaster/cards/bellordpack/Midnight.java
```diff
@@ -0,0 +1,40 @@
...
+
+public class Midnight extends AbstractPackmasterCard implements OnObtainCard {
+    public final static String ID = makeID("Midnight");
+    // intellij stuff attack, enemy, common, 9, 3, 9, 3, , 
+
+    public Midnight() {
+        super(ID, 1, CardType.ATTACK, CardRarity.COMMON, CardTarget.ENEMY);
+        baseDamage = 9;
+        baseBlock = 9;
+        cardsToPreview = new CurseOfTheBell();
```
erasels, line 23: Does it need to preview the card after it is already obtained? Not like it has any further effect on the card. Might be neat to remove it after pickup.

PR #99, opened 2023-01-10
Eurogame Addition 2

NoMasPollo:
New version with some balance and efficiency changes, still hopefully no bugs.


src/main/java/thePackmaster/SpireAnniversary5Mod.java
```diff
@@ -378,6 +378,9 @@ public void receiveAddAudio() {
     @Override
     public void receiveOnBattleStart(AbstractRoom room) {
         pandaList.clear();
+        if (p().masterDeck.contains(new Auctioneering()) || p().masterDeck.contains(new Harvest()) || p().masterDeck.contains(new NetFishing()) || p().masterDeck.contains(new IronMining()) || p().masterDeck.contains(new LumberFelling())) {
```
erasels, line 381: I don't remember how contains checks for equality. Is a newly created instance equal?
If it is, you don't have to change this, just a small tip. You could've created an interface `VictoryCard` (and implemented it on these cards) and just iterated through your deck and checked if any card was instanceOf that interface

NoMasPollo: ok gonna see about fixing that

PR #107, opened 2023-01-10
Fixed False Grit's upgrade not displaying in compendium

ActuallyEnbeon:
Assuming no more bugs appear, this is probably my last PR (at least until after release). Thanks Gk for all your hard work ^^


ActuallyEnbeon:
Turns out there are a couple of other changes I should make so I'm closing this, will re-PR later.

PR #109, opened 2023-01-11
Changes

NoMasPollo:
Significant nerfs to passive VP gain from non-pack cards, no longer removes stacks when you are already in the stance. Nerfs and buffs to a few cards


NoMasPollo:
It also makes calculations made by the player easier, however

PR #124, opened 2023-01-12
Added Graveyard pack, with images. Made mod listen for exhausts.

Charlie-Steiner:
Added graveyard pack and images. Everything is self-contained except a "combatExhausts" variable in the mod.java to track how many cards have been exhausted in-combat.


src/main/java/thePackmaster/actions/graveyardpack/ExhaustFromHandAction.java
```diff
@@ -0,0 +1,151 @@
...
+import com.megacrit.cardcrawl.actions.common.GainEnergyAction;
+import com.megacrit.cardcrawl.cards.AbstractCard;
+import com.megacrit.cardcrawl.characters.AbstractPlayer;
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.core.CardCrawlGame;
+import com.megacrit.cardcrawl.core.Settings;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.localization.UIStrings;
+
+public class ExhaustFromHandAction extends AbstractGameAction {
```
erasels, line 14: This action is a bit of a mess. If I understand it correctly, you just want to exhaust cards from hand without a screen.

This might be more useful for a full character but since it's only used once in this character, I'd rather not deal with it. All that card does is exhaust all cards in hand, so please use an anonymous action `new AbstractGameAction() {...` call ExhaustSpecificCard action for every card in hand which is how base game does it.

src/main/java/thePackmaster/actions/graveyardpack/PlayFromExhaustAction.java
```diff
@@ -0,0 +1,156 @@
...
+  	    	for (Iterator<AbstractCard> ex = this.p.exhaustPile.group.iterator(); ex.hasNext(); ) {
+  		        AbstractCard derp = (AbstractCard)ex.next();
+  		        if (derp.color.equals(AbstractCard.CardColor.COLORLESS) || derp.color.equals(AbstractCard.CardColor.CURSE)) {
+  		          this.classless.add(derp);
+  		          ex.remove();
+  		        } 
+  		    }
+  	    	
+  	      if (this.p.exhaustPile.isEmpty()) {
+  	    	  this.p.exhaustPile.group.addAll(classless);
```
erasels, line 61: If the pile is empty, add everything from it? I think there's a problem here.

src/main/java/thePackmaster/actions/graveyardpack/PlayFromExhaustAction.java
```diff
@@ -0,0 +1,156 @@
...
+import thePackmaster.actions.graveyardpack.CleanUpCardAction;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+
+
+public class PlayFromExhaustAction extends AbstractGameAction
```
erasels, line 30: This is another action that's a bit of a nightmare to read through. Please do another pass on it

src/main/java/thePackmaster/actions/graveyardpack/PlayRandomFromExhaustAction.java
```diff
@@ -0,0 +1,122 @@
...
+import com.megacrit.cardcrawl.vfx.cardManip.ShowCardBrieflyEffect;
+
+import thePackmaster.actions.graveyardpack.CleanUpCardAction;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+
+
+
+public class PlayRandomFromExhaustAction extends AbstractGameAction
```
erasels, line 24: I'm sure this can be combined with the PlayFromExhaustAction

src/main/java/thePackmaster/actions/graveyardpack/ReturnExhaustedToDeckAction.java
```diff
@@ -0,0 +1,141 @@
...
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.localization.UIStrings;
+import com.megacrit.cardcrawl.relics.AbstractRelic;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class ReturnExhaustedToDeckAction
```
erasels, line 19: I haven't looked at this, I'll just ask if this can use either `ExhaustToHandAction` (takes a card) or `ExhumeAction` (like exhume) instead of making this

src/main/java/thePackmaster/cards/graveyardpack/DarkOfNight.java
```diff
@@ -0,0 +1,45 @@
...
+	public DarkOfNight() {
+		super(ID, COST, TYPE, RARITY, TARGET);
+		
+		this.baseMagicNumber=1;
+		this.magicNumber=this.baseMagicNumber;
+	}
+
+  public void use(AbstractPlayer p, AbstractMonster m) {
+      AbstractDungeon.actionManager.addToBottom(new ApplyPowerAction(p, p, new DarkOfNightPower(p,this.magicNumber), this.magicNumber));
+      CardCrawlGame.sound.play("EVENT_SPIRITS", 0.05F);
```
erasels, line 38: Please use the SFX action

src/main/java/thePackmaster/cards/graveyardpack/ExtraLimbs.java
```diff
@@ -0,0 +1,60 @@
...
+		for (int x = 0; x < this.magicNumber; x++) {
+			AbstractDungeon.actionManager
+					.addToBottom(new DamageAction(m, new DamageInfo(p, this.damage, this.damageTypeForTurn),
+							AbstractGameAction.AttackEffect.SLASH_DIAGONAL));
+		}
+	}
+
+	
+	public void triggerOnExhaust() {
+		this.baseMagicNumber++;
```
erasels, line 53: Pretty sure you need to modify magicNumber not baseMagic

src/main/java/thePackmaster/cards/graveyardpack/FuneraryJewels.java
```diff
@@ -0,0 +1,49 @@
...
+import com.megacrit.cardcrawl.monsters.AbstractMonster;
+import com.megacrit.cardcrawl.powers.DexterityPower;
+import com.megacrit.cardcrawl.powers.StrengthPower;
+import com.megacrit.cardcrawl.actions.common.ApplyPowerAction;
+import com.megacrit.cardcrawl.actions.common.DrawCardAction;
+import com.megacrit.cardcrawl.actions.common.GainBlockAction;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class FuneraryJewels
```
erasels, line 17: This card seems kind of bad. Could probably stand to increase block a little or do something else with it

src/main/java/thePackmaster/cards/graveyardpack/LastGasp.java
```diff
@@ -0,0 +1,47 @@
...
+
+  
+	public LastGasp() {
+		super(ID, COST, TYPE, RARITY, TARGET);
+		
+  		this.isEthereal=true;
+	}
+
+  public void use(AbstractPlayer p, AbstractMonster m) {
+	  AbstractDungeon.actionManager.addToTop(new PlayFromExhaustAction());
```
erasels, line 40:
```suggestion
	  AbstractDungeon.actionManager.addToBot(new PlayFromExhaustAction());
```

src/main/java/thePackmaster/cards/graveyardpack/SpontaneousRitual.java
```diff
@@ -0,0 +1,132 @@
...
+		this.baseMagicNumber=0;
+		this.magicNumber=this.baseMagicNumber;
+		
+	}
+
+	  public void applyPowers() {
+	    super.applyPowers();
+	    
+	    this.baseDamage=SpireAnniversary5Mod.combatExhausts;
+	    this.damage=this.baseDamage;
```
erasels, line 48: This means that the damage won't be modified by strength, etc. Are you sure you want that?

src/main/java/thePackmaster/cards/graveyardpack/SpontaneousRitual.java
```diff
@@ -0,0 +1,132 @@
...
+
+	    
+		if(this.upgraded) {
+				tempDamage+= 3 - java.lang.Math.max(p.hand.size()-BaseMod.MAX_HAND_SIZE+2, 0);	//when hand is too large, don't add the full 3 to damage (extra wounds are automatically discarded)
+				AbstractDungeon.actionManager.addToTop(new MakeTempCardInHandAction(new Dazed(), 3));
+		}
+		
+		AbstractDungeon.actionManager.addToBottom(new ExhaustFromHandAction(BaseMod.MAX_HAND_SIZE,true,false,false));	//p.hand.size() not updating correctly for some reason
+
+	    CardCrawlGame.sound.play("ORB_DARK_EVOKE", 0.1F);
```
erasels, line 82: SFX Action

src/main/java/thePackmaster/cards/graveyardpack/SpontaneousRitual.java
```diff
@@ -0,0 +1,132 @@
...
+	    
+	    
+		DamageInfo d = new DamageInfo(p, tempDamage, this.damageTypeForTurn);
+		d.applyPowers(p, m);
+		
+	    AbstractDungeon.actionManager.addToBottom(new DamageAction(m, d, AbstractGameAction.AttackEffect.FIRE));
+
+	    
+	    if(AbstractDungeon.player.hasRelic("Unceasing Top"))	//hack to get around unceasing top not triggering correctly
```
erasels, line 91: Please use UnceasingTop.ID
This like is only needed because the action that exhausts the card doesn't correctly call refreshHandLayout()

So likely won't be needed once you fix that

src/main/resources/anniv5Resources/localization/eng/graveyardpack/Cardstrings.json
```diff
@@ -0,0 +1,49 @@
+{
+  "${ModID}:GreyBargain": {
+    "NAME": "Grey Bargain",
+    "DESCRIPTION": "Return a ${ModID}:Buried card to your draw pile by Exhausting a card in your hand.",
+    "UPGRADE_DESCRIPTION": "Return a ${ModID}:Buried card to your draw pile by Exhausting a card in your hand. NL Draw a card.",
+    "EXTENDED_DESCRIPTION": ["I need a card to Exhaust."]
+  },
+    "${ModID}:ExtraLimbs": {
+    "NAME": "Extra Limbs",
+    "DESCRIPTION": "Deal !D! damage !M! times. NL When Exhausted, increase the number of strikes by 1 this combat. NL Exhaust."
```
erasels, line 10:
```suggestion
    "DESCRIPTION": "Deal !D! damage !M! times. NL When Exhausted, increase the number of hits by 1 this combat. NL Exhaust."
```

src/main/resources/anniv5Resources/localization/eng/graveyardpack/Keywordstrings.json
```diff
@@ -0,0 +1,17 @@
+[
+  {
+    "PROPER_NAME": "Buried",
+    "NAMES": [
+      "buried"
+    ],
+    "DESCRIPTION": "A class card (not Colorless, Status, or Curse) from your exhaust pile."
+  },
+  {
+    "PROPER_NAME": "Purge",
```
erasels, line 10: Already a keyword in stslib

src/main/resources/anniv5Resources/localization/eng/graveyardpack/Cardstrings.json
```diff
@@ -0,0 +1,49 @@
...
+    "NAME": "Ancient Spear",
+    "DESCRIPTION": "Deal !D! damage. NL Exhaust."
+  },
+    "${ModID}:FleshLikeOak": {
+    "NAME": "Flesh Like Oak",
+    "DESCRIPTION": "Gain !B! Block. NL Gain !M! Strength. NL Exhaust."
+  },
+    "${ModID}:CallOfTheGrave": {
+    "NAME": "Call of the Grave",
+    "DESCRIPTION": "At the start of your turn, play a random ${ModID}:Buried card and ${ModID}:Purge it.",
```
erasels, line 22: Purge exists in stslib

src/main/java/thePackmaster/cards/graveyardpack/SpontaneousRitual.java
```diff
@@ -0,0 +1,132 @@
...
+	private static final int COST = 1;
+
+	public SpontaneousRitual() {
+		super(ID, COST, TYPE, RARITY, TARGET);
+
+		this.baseDamage = 0;
+		this.damage = this.baseDamage;
+		this.baseMagicNumber=0;
+		this.magicNumber=this.baseMagicNumber;
+
```
erasels, line 41:
```suggestion
		cardsToPreview = new Dazed();
```

erasels (changes requested):

src/main/java/thePackmaster/actions/graveyardpack/PlayFromExhaustAction.java
Charlie-Steiner: Not a bug. This adds things back into the exhaust pile - it seems odd, but I think it's the straightforward way to use the original game's method for picking cards from the exhaust without letting the player pick colorless cards.

src/main/java/thePackmaster/actions/graveyardpack/PlayRandomFromExhaustAction.java
Charlie-Steiner: One has to open up UI for the player and the other doesn't, so I'd rather keep them separate.

src/main/java/thePackmaster/actions/graveyardpack/PlayFromExhaustAction.java
erasels: Oh I see, I got confused by the different piles

src/main/java/thePackmaster/cards/graveyardpack/ExtraLimbs.java
Charlie-Steiner: Works in testing.

src/main/java/thePackmaster/cards/graveyardpack/FuneraryJewels.java
Charlie-Steiner: I guess I could change it to the slightly more busted version that costs 1 and refunds you 1 energy.

src/main/java/thePackmaster/cards/graveyardpack/SpontaneousRitual.java
Charlie-Steiner: Ahah, this bug was bugging me! Awesome.

src/main/java/thePackmaster/cards/graveyardpack/SpontaneousRitual.java
Charlie-Steiner: Applies strength in testing.

src/main/java/thePackmaster/actions/graveyardpack/ReturnExhaustedToDeckAction.java
Charlie-Steiner: I think this needs to be custom, both because ExhumeAction really wants to put the card in your hand, and to exclude non-class cards.

Charlie-Steiner: Nevermind, I replaced it with MoveCardsAction plus a shuffle.

src/main/java/thePackmaster/actions/graveyardpack/PlayRandomFromExhaustAction.java
Charlie-Steiner: I decided to merge them after all.

erasels (approved):

erasels:
Alright, after deliberating I just can't merge spontaneous Ritual in its current state. This card has some issues, even if it works as intended.

It does not show the amount of damage it will deal, only the amount of cards exhausted, so the player can#t know how pen nib affects it or vuln, etc. If you were to show the damage on the card, it would not work because right now you're not applying powers to the card correctly, it still works because calculateCardDamage (basically applyPowers but when hovering a monster with the card targeting reticle) and applying powers to the damage info in the use method.

Please correctly use damage by setting baseDamage to cards exhausted before calling super.applyPowers (you can also add your pre-add-dazed logic here). This way damage will show the correctly modified cardsExhausted+multipliers damage when the card is in hand. This way you can also pass damage to the damageInfo in use instead of calling damageInfo.applyPowers on it.
You'll also have to basically copy this logic to calculateCardDamage where you'll replace the super applyPowers call with super calculateCardDamage

PR #134, opened 2023-01-13
Claw Pack!

mikemayhemdev:


src/main/java/thePackmaster/SpireAnniversary5Mod.java
```diff
@@ -698,4 +713,23 @@ public void receiveStartGame() {
         }
     }
 
+    public static void addPotions() {
+
+        BaseMod.addPotion(AttackPotionButClaw.class, Color.RED, Color.WHITE, Color.FIREBRICK, AttackPotionButClaw.POTION_ID);
+        BaseMod.addPotion(ClawPowerPotion.class, Color.RED, Color.WHITE, Color.FIREBRICK, ClawPowerPotion.POTION_ID);
+        BaseMod.addPotion(DrawClawsPotion.class, Color.RED, Color.WHITE, Color.FIREBRICK, DrawClawsPotion.POTION_ID);
+        BaseMod.addPotion(GenerateClawsPotion.class, Color.RED, Color.WHITE, Color.FIREBRICK, GenerateClawsPotion.POTION_ID);
```
modargo, line 721: Currently this will make these potions show up for all characters. Need to pass a final argument of the character enum, `ThePackmaster.Enums.THE_PACKMASTER`.

And since these are specific to the claw pack, you should override `getPackPotions` in the `ClawPack` class to return a list of the potion IDs.

src/main/java/thePackmaster/cardmodifiers/strikepack/AddStrikeTagModifier.java
```diff
@@ -15,15 +17,16 @@ public void onInitialApplication(AbstractCard card) {
...
+            card.applyPowers();
         }
     }
 
     @Override
-    public void alterName(AbstractCard card) {
-        modifyName(card.name + " Strike", card);
-        super.alterName(card);
+    public String nameSuffix(String cardName) {
+        return " Strike";
```
modargo, line 26: This should probably be taken from a JSON file and not hardcoded, right? Translating this mod seems like quite a task, but still best to avoid hardcoded strings.

For miscellaneous strings like this I usually put them in the UIStrings file for the pack.

src/main/java/thePackmaster/cardmodifiers/clawpack/AddClawTagAndMakeClawModifier.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    public void onUse(AbstractCard card, AbstractCreature target, UseCardAction action) {
+        super.onUse(card, target, action);
+        AbstractCard c = new GhostClaw();
+        Wiz.atb(new MakeTempCardInHandAction(c));
+    }
+
+    @Override
+    public String nameSuffix(String cardName) {
+        return " Claw";
```
modargo, line 52: Same comment about not hardcoding strings.

src/main/java/thePackmaster/cards/clawpack/TwinClaw.java
```diff
@@ -0,0 +1,46 @@
...
+            this.addToBot(new VFXAction(new ClawEffect(m.hb.cX, m.hb.cY, Color.RED, Color.WHITE), 0.1F));
+        }
+        this.addToBot(new DamageAction(m, new DamageInfo(p, this.damage, DamageInfo.DamageType.NORMAL), AbstractGameAction.AttackEffect.NONE));
+        this.addToBot(new DamageAction(m, new DamageInfo(p, this.damage, DamageInfo.DamageType.NORMAL), AbstractGameAction.AttackEffect.NONE));
+
+        ClawUp(magicNumber);
+    }
+
+    public void upp() {
+        upgradeMagicNumber(2);
```
modargo, line 44: Is upgrading magic number instead of damage for Twin Claw intentional? I ask because I was expecting this to get +2 damage on upgrade like Twin Strike does, instead of increasing claw scaling.

Totally fine if that's how you want the card to work, just commenting in case this was an oversight.

src/main/java/thePackmaster/cards/cthulhupack/BeyondTheStars.java
```diff
@@ -13,20 +17,20 @@ public class BeyondTheStars extends AbstractCthulhuCard {
...
+        cardsToPreview = new Madness();
     }
 
 
     public void use(AbstractPlayer p, AbstractMonster m) {
         allDmg(AbstractGameAction.AttackEffect.FIRE);
-        Wiz.applyToSelf(new EntropyPower(p, this.magicNumber));
+        AbstractCard c = new Madness();
+        if (upgraded) c.upgrade();
+        AbstractDungeon.actionManager.addToBottom(new MakeTempCardInHandAction(c.makeCopy(), 1));
```
modargo, line 30: Won't break anything, but when you've just made a card using `new`, immediately calling `makeCopy` isn't needed.

modargo:
While you're changing things in the Dimension Gate pack, maybe also nerf Knighthood -- increase its cost and/or make it Ethereal?

This is what I posted in Discord two days ago (slightly edited for readability), don't know if you saw it:

> One piece of feedback on your Dimension Gate pack, I think Knighthood could use a nerf. Permanently upgrading is a very strong effect and the "card in hand" restriction is much easier to meet than fatal restriction of Lesson Learned (which is already a great card and the closest base game comparison for Knighthood). This is based on my play experience with it -- I got Knighthood in one of my runs yesterday and felt it was too easy to use for how much benefit it provided.

> My suggestion would be to at least make it cost 3 (like Wish, another card that gives a permanent benefit). Or consider some other nerf; my intuition is that even costing 3 it will be a strong card that might border on too strong. (It also encourages stalling, which isn't great but is inherent to cards like Knighthood and Wish, so I'm okay with that aspect of the card.)

Gk then suggested Ethereal to make it more awkward to play, which seemed like a good idea to me.

I know this is tangential to the main point of the PR, but thought it was worth mentioning. Feel free to postpone this discussion to some other time if you think this needs more consideration/discussion and don't want to derail things.

src/main/java/thePackmaster/cards/dimensiongatepack/Inferno.java
```diff
@@ -1,43 +1,65 @@
...
-        exhaust = true;
+        super(ID, -1, CardRarity.RARE, AbstractCard.CardType.ATTACK, AbstractCard.CardTarget.ALL);
+        baseDamage = 100;
         setFrame("infernoframe.png");
         isMultiDamage = true;
-
+        PersistFields.basePersist.set(this, 2);
+        PersistFields.persist.set(this, 2);
+        HoardField.baseHoard.set(this, 9);
+        HoardField.hoard.set(this, 9);
```
modargo, line 29: Interesting to see where this ended up... reasonably translating this effect from Monster Train seems hard.

src/main/java/thePackmaster/SpireAnniversary5Mod.java
modargo: (The functionality for getPackPotions doesn't fully work yet, but Pandemonium is working on it.)

modargo: After looking at the potions themselves I see they use the Placeholder rarity... which prevents some issues but not all of them, since there are sources of potions that don't respect rarity. Specifically anything that uses `PotionHelper.getRandomPotion`, which returns a uniformly random potion -- so the events Knowing Skull, Lab, Woman in Blue, the relics Cauldron, Tiny House, and probably one or two other things.

I've done something like this for Corrupt the Spire and I prevented the potions from showing up by patching that method. We should do something similar here (in addition to marking these as Packmaster only potions, good to be safe). Here's the patch from Corrupt the Spire if you want to adapt it: https://github.com/modargo/corruptthespire/blob/f47f2a7b9ee5950e128d01960d519c6ed949fb7d/src/main/java/corruptthespire/patches/potions/FilterOutCorruptedPotionsPatch.java

modargo: Also, if the idea is that tehse are a special rarity and don't normally show up (i.e. they only show up from Alclawmize), you can ignore what I said about getPackPotions. Sorry for any confusion.

modargo:
I took a pass through this to take some of the burden of PR reviews off Gk. Hope the comments all make sense, probably good for Gk or me to do a second pass after any changes based on the comments I've made.

modargo:
P.S.: I really look forward to playing with this pack. Hard to judge where the balance ends up but there's enough Claw generation and scaling that it could be a strong way to do damage, and I'm excited to experience how it plays. If it ends up being too strong, it'll be pretty funny anyway.

src/main/java/thePackmaster/SpireAnniversary5Mod.java
mikemayhemdev: These are indeed completely exclusive to Alclawmize.  But it sounds like I still will need that patch.

src/main/java/thePackmaster/cards/clawpack/TwinClaw.java
mikemayhemdev: Yeah this was intended, as the idea being that by default, if it was truly "two claws" it should be +4 sharp but that'd be too strong of a card by base.  So when upgraded its actually just two claws.

src/main/java/thePackmaster/patches/clawpack/ClawUsePatch.java
```diff
@@ -0,0 +1,25 @@
...
+import thePackmaster.util.Wiz;
+
+@SpirePatch(
+        clz = GashAction.class,
+        method = "update"
+)
+public class ClawUsePatch {
+        public static SpireReturn Postfix(GashAction __instance) {
+            Wiz.applyToSelf(new ClawStrengthPower(Wiz.p(), __instance.amount));
+            return SpireReturn.Return();
```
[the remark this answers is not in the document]
mikemayhemdev: Okay, I ended up going with something very similar to this.  GashAction is prefix'd to go also run an identical process on any CLAW-tagged cards, excepting actual Claw instances since they are covered by the Gash Action.  Other cards that do Sharpen just run the custom version and does not except Claws.  This is necessary because the GashAction receives a card input and modifies it directly, and not all the sources of Sharpen are specifically on a card source.

Additionally the Ghost Claw now inherits previous Sharpen effects, which is explained on the keyword for Sharpen.

ClawStrengthPower is dead.

src/main/java/thePackmaster/cards/clawpack/AbstractClawCard.java
```diff
@@ -17,10 +21,64 @@ public AbstractClawCard(String cardID, int cost, CardType type, CardRarity rarit
...
+            if (!(excludeClaws && (c instanceof Claw)) && c.hasTag(SpireAnniversary5Mod.CLAW)) {
+                c.baseDamage += value;
+                c.applyPowers();
+                if (c.cardsToPreview != null){
+                    if (c.cardsToPreview instanceof GhostClaw){
+                        ((GhostClaw)c).refreshDamage();
+                    }
+                }
+            }
+        }
```
modargo, line 81: I can see that you took this from the decompiled GashAction (thus all the iterators). I'm sure it works just fine, but this can be written in a way that's a lot cleaner, and it's a pretty easy change to make. I think I can make a Github suggestion for code to replace this, so try the below on for size! It's much more readable.

```suggestion
         List<AbstractCard> cards = new ArrayList<>();
        cards.addAll(AbstractDungeon.player.hand.group);
        cards.addAll(AbstractDungeon.player.drawPile.group);
        cards.addAll(AbstractDungeon.player.discardPile.group);
        
        for (AbstractCard c : cards) {
            if (!(excludeClaws && (c instanceof Claw)) && c.hasTag(SpireAnniversary5Mod.CLAW)) {
                c.baseDamage += value;
                c.applyPowers();
                if (c.cardsToPreview != null){
                    if (c.cardsToPreview instanceof GhostClaw){
                        ((GhostClaw)c).refreshDamage();
                    }
                }
            }
        }
```

src/main/java/thePackmaster/cards/dimensiongatepack/Knighthood.java
```diff
@@ -21,6 +21,7 @@ public Knighthood() {
         setFrame("knighthoodframe.png");
         exhaust = true;
         tags.add(CardTags.HEALING);
+        isEthereal = true;
```
modargo, line 24: Yay! Makes it a bit harder to use, can't stall with it unless you have discard or exhume effects. That's a big step towards making it a more reasonable card, and we'll see how it plays.

src/main/java/thePackmaster/patches/FilterOutPlaceholderPotionsPatch.java
```diff
@@ -0,0 +1,37 @@
...
+import com.megacrit.cardcrawl.potions.AbstractPotion;
+import com.megacrit.cardcrawl.random.Random;
+import javassist.CannotCompileException;
+import javassist.expr.ExprEditor;
+import javassist.expr.FieldAccess;
+
+import java.util.ArrayList;
+import java.util.stream.Collectors;
+
+public class FilterOutPlaceholderPotionsPatch {
```
modargo, line 15: Makes sense, good to see the sample patch I linked to was useful.

src/main/java/thePackmaster/patches/clawpack/ClawUsePatch.java
```diff
@@ -10,16 +12,20 @@
...
 )
 public class ClawUsePatch {
-        public static SpireReturn Postfix(GashAction __instance) {
-            Wiz.applyToSelf(new ClawStrengthPower(Wiz.p(), __instance.amount));
-            return SpireReturn.Return();
-        }
+
+    @SpirePrefixPatch
+    public static void Prefix(GashAction __instance) {
+        AbstractClawCard.ClawUp(__instance.amount, true);
```
modargo, line 28: Yeah, this seems like a fine way to do it. The version that doesn't touch base game Claws is called here (since GashAction already handles that), while the version that does touch base game Claws is called everywhere else.

Glad we made this change.

modargo:
The core set art is really cool. Awesome to see it.

Backpack Smack is absolutely hilarious.

src/main/java/thePackmaster/SpireAnniversary5Mod.java
```diff
@@ -324,6 +326,26 @@ public String glowID() {
         addPotions();
     }
 
+
+    public static void addPotions() {
+
+        BaseMod.addPotion(AttackPotionButClaw.class, Color.RED, Color.WHITE, Color.FIREBRICK, AttackPotionButClaw.POTION_ID);
+        BaseMod.addPotion(ClawPowerPotion.class, Color.RED, Color.WHITE, Color.FIREBRICK, ClawPowerPotion.POTION_ID);
+        BaseMod.addPotion(DrawClawsPotion.class, Color.RED, Color.WHITE, Color.FIREBRICK, DrawClawsPotion.POTION_ID);
+        BaseMod.addPotion(GenerateClawsPotion.class, Color.RED, Color.WHITE, Color.FIREBRICK, GenerateClawsPotion.POTION_ID);
```
modargo, line 335: Let's add in the packmaster class enum as the final argument here. With how these are only generated by Alclawmize, I know that it doesn't really matter for these specific potions -- but my concern is more about what happens when the _next_ person adds a custom packmaster potion. I want them to see that every other potion is specific to the packmaster, and then make their potion specific to the packmaster too (they might even just copy paste one of these lines).

Going to do another Github suggestion here, I think there's a way you can just accept these and it makes the change.

```suggestion
        BaseMod.addPotion(AttackPotionButClaw.class, Color.RED, Color.WHITE, Color.FIREBRICK, AttackPotionButClaw.POTION_ID, ThePackmaster.Enums.THE_PACKMASTER);
        BaseMod.addPotion(ClawPowerPotion.class, Color.RED, Color.WHITE, Color.FIREBRICK, ClawPowerPotion.POTION_ID, ThePackmaster.Enums.THE_PACKMASTER);
        BaseMod.addPotion(DrawClawsPotion.class, Color.RED, Color.WHITE, Color.FIREBRICK, DrawClawsPotion.POTION_ID, ThePackmaster.Enums.THE_PACKMASTER);
        BaseMod.addPotion(GenerateClawsPotion.class, Color.RED, Color.WHITE, Color.FIREBRICK, GenerateClawsPotion.POTION_ID, ThePackmaster.Enums.THE_PACKMASTER);
```

src/main/java/thePackmaster/SpireAnniversary5Mod.java
```diff
@@ -324,6 +326,26 @@ public String glowID() {
...
+
+        /*
+        //TODO - Just before ship, add this cause otherwise we'll trigger a dependency for everyone.
```
modargo, line 338: As a separate follow up, I have a way to do this wide potions stuff that does not require any build dependency at all (and I think we should avoid build dependencies to make it as easy as possible for contributors to build this). After all, it sounds like there are people who want to contribute packs after release, and I'm sure even people who have finished their packs will want to tweak things.

So no need to worry about it for now, I'll make a future PR that changes this.

src/main/java/thePackmaster/cards/clawpack/AbstractClawCard.java
modargo: I also noticed a bug in this code! For the Ghost Claw check:
```
                    if (c.cardsToPreview instanceof GhostClaw){
                        ((GhostClaw)c).refreshDamage();
                    }
```

You're checking whether `c.cardsToPreview` is a Ghost Claw, but then you're trying to cast `c`. This will crash when you use the attack potion but Claw.

I've edited my suggestion for making this code more readable to also include a fix for this.

src/main/java/thePackmaster/cards/Rummage.java
```diff
@@ -28,19 +28,29 @@ public void use(AbstractPlayer p, AbstractMonster m) {
...
-                for (AbstractCard q : AbstractDungeon.player.hand.group) {
-                    if (Wiz.getLogicalCardCost(q) > 0)
-                        possCardsList.add(q);
-                }
+                ArrayList<AbstractCard> chosenCards = new ArrayList<>();
+                int highestcost = 0;
 
                 for (int i = 0; i < magicNumber; i++) {
+                    for (AbstractCard q : AbstractDungeon.player.hand.group) {
+                        if (!chosenCards.contains(q)) {
```
modargo, line 36: There's a bug here with X-cost cards. Cost reduction is meaningless for X-cost cards, but nothing in this code filters them out, and `Wiz.getLogicalCardCost` returns your current energy total for an X-cost card.

The fix is to filter out those cards on this line:
```suggestion
                        if (!chosenCards.contains(q) && q.cost > 0) {
```

Checking `q.cost > 0` will exclude three categories of cards that this code should avoid: unplayable cards, X-cost cards, and 0-cost cards. In all those cases the cost reduction is meaningless but Rummage will still act like its reducing their cost.

src/main/java/thePackmaster/cards/clawpack/SearingClaw.java
```diff
@@ -0,0 +1,56 @@
...
+        super(ID, 2, CardType.ATTACK, CardRarity.UNCOMMON, CardTarget.ENEMY);
+        baseDamage = 8;
+        baseMagicNumber = magicNumber = 2;
+        tags.add(CLAW);
+
+    }
+
+    public void use(AbstractPlayer p, AbstractMonster m) {
+
+        this.addToBot(new VFXAction(new ClawEffect(m.hb.cX, m.hb.cY, Color.ORANGE, Color.WHITE), 0.1F));
```
modargo, line 31: This should totally do one claw effect for each time this card has been upgraded 😃.

src/main/resources/anniv5Resources/localization/eng/dimensiongatepack/Cardstrings.json
```diff
@@ -19,14 +19,14 @@
   },
   "${ModID}:Knighthood": {
     "NAME": "Knighthood",
-    "DESCRIPTION": "Choose a card in hand to Upgrade permanently. NL Exhaust.",
+    "DESCRIPTION": "Ethereal. NL Choose a card in hand to Upgrade permanently. NL Exhaust.",
```
modargo, line 22: With the new Ethereal -> not Ethereal upgrade, Knighthood needs an `UPGRADE_DESCRIPTION` entry added.

mikemayhemdev:
Alright, I completed the last comments from modargo, as well as simplified the issues surrounding base Claw by just making a PMClaw duplicate.  This means the only time the GashAction override ever triggers is if you acquire a true prismatic shard Claw, but still works as it did before - I made this change because of the Sharpen keyword mentioning Ghost Claw by name, and just existing in general.  The PMClaw still looks like the Defect card, but reads "Sharpen 2" instead of the normal claw bonus card text.

I think we should be ready to roll now.

src/main/java/thePackmaster/cards/clawpack/Conclawd.java
```diff
@@ -0,0 +1,45 @@
...
+
+public class Conclawd extends AbstractClawCard {
+    public final static String ID = makeID("Conclawd");
+
+    public Conclawd() {
+        super(ID, 1, CardType.ATTACK, CardRarity.UNCOMMON, CardTarget.ALL_ENEMY);
+        baseDamage = 10;
+        baseMagicNumber = magicNumber = 2;
+        tags.add(CLAW);
+
```
erasels, line 26:
```suggestion
isMultiDamage = true;
```

src/main/java/thePackmaster/cards/clawpack/PMClaw.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class PMClaw extends AbstractClawCard {
+    public final static String ID = makeID("PMClaw");
+
+    public PMClaw() {
+        super(ID, 0, CardType.ATTACK, CardRarity.COMMON, CardTarget.ENEMY, CardColor.BLUE);
+        //Does this do a Card Art Roller first, then override it?  Probably.  Might be worth having a 'skip art roller' bool?
+        //Seems better than needing to copy Claw's base art as a new file.
+        TextureAtlas atlas = ReflectionHacks.getPrivate(this, AbstractCard.class, "cardAtlas");
+        this.portrait = atlas.findRegion("blue/attack/claw");
```
erasels, line 28: roller has this functionality

https://github.com/erasels/PackmasterCharacter/blob/main/src/main/java/thePackmaster/cards/WitchesStrike/ChitteringPunt.java#L39

src/main/java/thePackmaster/cards/dimensiongatepack/Inferno.java
```diff
@@ -1,43 +1,65 @@
...
     public void use(AbstractPlayer p, AbstractMonster m) {
-        allDmg(AbstractGameAction.AttackEffect.FIRE);
-        addToBot(new AbstractGameAction() {
-            @Override
-            public void update() {
-                if (!AbstractDungeon.getCurrRoom().monsters.areMonstersBasicallyDead()){
-                    addToBot(new SelfDamageAction(new DamageInfo(p, damage, DamageInfo.DamageType.NORMAL), AbstractGameAction.AttackEffect.FIRE));
-                }
-                this.isDone = true;
+        BaseMod.logger.info("Inferno's MN start: " + HoardField.hoard.get(this));
```
erasels, line 35: make this an action

src/main/resources/anniv5Resources/localization/eng/clawpack/Keywordstrings.json
```diff
@@ -0,0 +1,9 @@
+[
+  {
+    "PROPER_NAME": "Sharpen",
+    "NAMES": [
+      "sharpen"
+    ],
+    "DESCRIPTION": "Raises the damage dealt by all existing Claw cards and any Ghost Claws created this combat."
```
erasels, line 7:
```suggestion
    "DESCRIPTION": "Raises the damage dealt by all existing #yClaws and any #yGhost #yClaws created this combat."
```

src/main/resources/anniv5Resources/localization/eng/clawpack/Powerstrings.json
```diff
@@ -0,0 +1,17 @@
...
+    "DESCRIPTIONS": [
+      "At the start of your turn, gain #b",
+      " #yGhost #yClaws.",
+      " #yGhost #yClaw."
+    ]
+  },
+  "${ModID}:ReturnBlockOnClawPower": {
+    "NAME": "Talk to the Claw",
+    "DESCRIPTIONS": [
+      "When attacked with a Claw, you gain #b",
```
erasels, line 13:
```suggestion
      "When attacked with a #yClaw, you gain #b",
```

erasels (changes requested):

erasels (approved):

PR #136, opened 2023-01-13
Make pack-exclusive potions be actually exclusive when registered as Packmaster potions

Pandemonium14:


modargo:
Nice how simple the final code ended up being. Thanks for doing this, it looks good to me.

PR #138, opened 2023-01-13
Rimworld Pack!

Laugic:
And also some art changes / wording updates


src/main/java/thePackmaster/powers/rimworldpack/DupePower.java
```diff
@@ -0,0 +1,59 @@
...
+import com.megacrit.cardcrawl.core.CardCrawlGame;
+import com.megacrit.cardcrawl.core.Settings;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.monsters.AbstractMonster;
+import thePackmaster.powers.AbstractPackmasterPower;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+import static thePackmaster.util.Wiz.adp;
+
+public class DupePower extends AbstractPackmasterPower {
```
modargo, line 19: Looks to me like the strings for this power are missing -- I don't see anything in Powerstrings.json for it.

I do understand why this exists and why you didn't use the base game power that Duplication Pot uses -- I'm guessing you don't want the part where it goes away at the end of the turn, since that's missing but the rest of the code is basically the same.

src/main/java/thePackmaster/powers/rimworldpack/MoodPower.java
```diff
@@ -0,0 +1,190 @@
...
+        amount += stackAmount;
+
+        checkForBreaksAndInsps();
+    }
+
+    private void checkForBreaksAndInsps(){
+
+        //NOTE: If I do an action to reduce or increase the amount instead of
+        //just doing it directly, I'd get a stack overflow if someone applies
+        //more than 4 mood at once (e.g by stacking copies of Sanguine)
```
modargo, line 67: Or by playing 4 copies of Burning Passion so all your buffs are multiplied by 16, then playing a Relax that normally adds 2 mood -> 32 mood.

src/main/java/thePackmaster/powers/rimworldpack/MoodPower.java
```diff
@@ -0,0 +1,190 @@
...
+                default:
+                    fast(owner);
+            }
+        }
+
+        private static void fast(AbstractCreature owner){
+
+            int rand = new Random().nextInt(10);
+            if(rand == 0)
+                Wiz.atb(new SFXAction(modID + "fast"));
```
modargo, line 137: So one time out of 10 you get an extra effect for this one.

Some people might find that annoying, but as an amusing little easter egg it's probably fine.

src/main/java/thePackmaster/powers/rimworldpack/DupePower.java
Laugic: Yup it was the end of turn effect that I didn't want. And also whoops, I have no idea how I missed the power strings for this

src/main/java/thePackmaster/powers/rimworldpack/MoodPower.java
Laugic: If you have 12 energy to spend on pure scaling you've basically already won the fight

src/main/java/thePackmaster/powers/rimworldpack/MoodPower.java
```diff
@@ -0,0 +1,190 @@
...
+                Wiz.atb(new VFXAction(new WhirlwindEffect(), 0.5f));
+                Wiz.atb(new VFXAction(new WhirlwindEffect(), 0.5f));
+            }
+        }
+
+        public static void mentalBreak(AbstractCreature owner)
+        {
+            int rand = AbstractDungeon.cardRandomRng.random(3) + 4;
+            Wiz.atb(new VFXAction(owner, new VerticalAuraEffect(new Color(1, 0, 0, .5f), owner.hb.cX, owner.hb.cY), 0.25F));
+            Wiz.atb(new TalkAction(true, inspirationStrings.TEXT[rand], 2.5f, 2.5f));
```
modargo, line 169: Instead of adding 4 here and subtracting 4 below, I'd suggest juts adding 4 here.

Or even better, this would probably be easier to manage if you split the inspirations and mental breaks into two separate entries in UIStrings. That will let you remove the random +- 4 here entirely, which is a good thing -- right now, if you ever add or remove an inspiration, it will also require you to change the number here. Easy to forget. I think it would also be easier to understand for any hypothetical translator (instead of having an entry that says `===Mental Breaks===` to separate the two).

Laugic: Sure thing, I can do that

src/main/java/thePackmaster/powers/rimworldpack/BurningPassionPower.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    public boolean onReceivePower(AbstractPower power, AbstractCreature target, AbstractCreature source) {
+        if(power.amount != 0 && target == owner && power.type == PowerType.BUFF)
+        {
+            flash();
+            power.amount *= amount;
+            power.updateDescription();
+        }
+        for (AbstractCard card: AbstractDungeon.player.hand.group)
```
modargo, line 55: I'm surprised this is needed, but even if it isn't needed, I guess it shouldn't do any harm.

src/main/java/thePackmaster/powers/rimworldpack/TradeBeaconPower.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public void atStartOfTurn() {
+        traded = false;
+        postStartDraw = false;
+    }
+
+    @Override
+    public void atStartOfTurnPostDraw() {
```
modargo, line 35: Per what we discussed in Discord, to make this consistently happen after any other effects that use this hook (most notably powers like DrawCardNextTurn, Brutality, and Tools of the Trade), this should either use priority or a doubly nested action to ensure it happens last.

modargo:
Looks good, I commented on the couple small things that could be cleaned up but I don't see any issues with the functionality for the cards themselves. I also checked out the PR locally and played around with the cards a bit, stacked Burning Passion up to some pretty high values but didn't try to achieve integer overflow.

I did not try to assess balance or anything since that's better done through playtesting.

Once you've made the small changes that I've suggested this should be ready to merge.

modargo:
Oh and just to record it here to make sure Gk/I remember when we next look at this PR, the one thing we discussed that isn't in a comment in this PR was adding a pack preview image (but it's fine for the rest of the art to come later).

Should be a fun pack to try out.

PR #139, opened 2023-01-13
fix PR

Dayvig:
fixed weirdness with carving knife. Hunted now decreases per turn, and skull club is no longer a common

### DIFF
--- a/src/main/java/thePackmaster/powers/monsterhunterpack/HuntersMark.java
+++ b/src/main/java/thePackmaster/powers/monsterhunterpack/HuntersMark.java
@@ -1,6 +1,7 @@
 package thePackmaster.powers.monsterhunterpack;
 
 import com.megacrit.cardcrawl.actions.common.GainBlockAction;
+import com.megacrit.cardcrawl.actions.common.ReducePowerAction;
 import com.megacrit.cardcrawl.actions.common.RemoveSpecificPowerAction;
 import com.megacrit.cardcrawl.cards.DamageInfo;
 import com.megacrit.cardcrawl.core.AbstractCreature;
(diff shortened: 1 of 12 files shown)


src/main/java/thePackmaster/powers/monsterhunterpack/HuntersMark.java
```diff
@@ -22,7 +22,7 @@ public class HuntersMark extends AbstractPackmasterPower {
     public static final String[] DESCRIPTIONS = powerStrings.DESCRIPTIONS;
 
     public HuntersMark(AbstractCreature owner, int amount) {
-        super(POWER_ID, LOC_NAME, PowerType.DEBUFF, false, owner, amount);
+        super(POWER_ID, LOC_NAME, PowerType.DEBUFF, true, owner, amount);
```
erasels, line 25: isTurnBased only changes the color of the number the power renders, it doesn't actually make the power turn based, you'll have to implement that logic yourself

src/main/java/thePackmaster/powers/monsterhunterpack/NotHunted.java
```diff
@@ -20,7 +20,7 @@ public class NotHunted extends AbstractPackmasterPower {
     public static final String[] DESCRIPTIONS = powerStrings.DESCRIPTIONS;
 
     public NotHunted(AbstractCreature owner, int amount) {
-        super(POWER_ID, LOC_NAME, PowerType.BUFF, false, owner, amount);
+        super(POWER_ID, LOC_NAME, PowerType.BUFF, true, owner, amount);
```
erasels, line 23: as above

erasels (changes requested):

PR #150, opened 2023-01-14
Energy and Echo | no art

Paul-Sml:
E&E with no art, will PR art later (before mod release).
Not against a code review, I've spent some time testing a lot of situations but I might have let some bugs slip.


Paul-Sml:
Ofc now there is art

src/main/java/thePackmaster/cards/energyandechopack/Slowing.java
```diff
@@ -0,0 +1,37 @@
...
+    public Slowing(){
+        super(ID, COST, CardType.POWER, CardRarity.UNCOMMON, CardTarget.SELF);
+        magicNumber = baseMagicNumber = 3;
+        cardsToPreview = new Wound();
+    }
+
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        applyToSelf(new BerserkPower(p, 1));
```
modargo, line 29: Do you want to add your own power for this instead of reusing Berserk?

Reusing Berserk does work for giving energy, but it will make the [E] icons show up as Ironclad energy icons (a bit weird) and the player will see a card named Slowing give a buff named Berserk (also weird).

Making a buff called Extra Energy or something like that, with some icon for it (maybe use the Packmaster's energy orb, or adapt the icon used by Charged Battery), would make the buff a lot more fitting.

src/main/java/thePackmaster/patches/warlockpack/NoFramePackPatch.java
```diff
@@ -0,0 +1,21 @@
...
+//import com.evacipated.cardcrawl.modthespire.lib.SpireReturn;
+//import com.megacrit.cardcrawl.cards.AbstractCard;
+//import thePackmaster.packs.WarlockPack;
+//
+//@SpirePatch(
+//        clz = AbstractCard.class,
+//        method = "renderSkillPortrait"
+//)
+//
+//public class NoFramePackPatch {
```
modargo, line 14: Should delete this file since it's all commented out. It can be recovered from history if needed -- or from this PR, for that matter :).

src/main/java/thePackmaster/powers/energyandechopack/ReceptorPower.java
```diff
@@ -0,0 +1,37 @@
...
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.core.CardCrawlGame;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.localization.PowerStrings;
+import com.megacrit.cardcrawl.powers.AbstractPower;
+import thePackmaster.powers.AbstractPackmasterPower;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class ReceptorPower extends AbstractPackmasterPower {
```
modargo, line 13: Is `ReceptorPower` an old version of DivineTouchPower? It looks like it's doing similar logic, but it doesn't have power strings or an icon for it, it doesn't look like anything applies it to the player. Should delete the file if that's the case.

src/main/java/thePackmaster/packs/EnergyAndEchoPack.java
```diff
@@ -0,0 +1,79 @@
...
+            @Override
+            public String glowID() {
+                return "anniv5:GlowEcho";
+            }
+        });
+
+        super.initializePack();
+    }
+
+    public void receiveOnBattleStart (AbstractRoom room) {
```
modargo, line 70: Pretty sure the issue you mentioned with energy used not resetting is because this is never called.

You can just move this call to the main mod file, in the `receiveOnBattleStart` method there -- there's already a couple similar things there.

src/main/java/thePackmaster/cardmodifiers/energyandechopack/EchoedEtherealMod.java
```diff
@@ -0,0 +1,42 @@
...
+import com.badlogic.gdx.graphics.Color;
+import com.megacrit.cardcrawl.cards.AbstractCard;
+import com.megacrit.cardcrawl.core.Settings;
+import com.megacrit.cardcrawl.helpers.GameDictionary;
+import com.megacrit.cardcrawl.localization.LocalizedStrings;
+import org.apache.commons.lang3.StringUtils;
+import thePackmaster.SpireAnniversary5Mod;
+import thePackmaster.cards.marisapack.AmplifyCard;
+
+//Basically Ethereal Modifier but it will go at the bottom and won't NL
```
modargo, line 15: Why have this be different from where the word Ethereal normally goes? Is it because you want to avoid adding new lines to cards? It looks to me like all the Echo cards have short enough descriptions that adding a line for Ethereal at the top would be okay. Though it looks like Tear Apart might reach six lines.

src/main/java/thePackmaster/cards/warlockpack/TheSoularium.java
```diff
@@ -39,6 +40,7 @@ public void use(AbstractPlayer p, AbstractMonster m) {
             public void update() {
                 for (AbstractCard c : DrawCardAction.drawnCards) {
                     CardModifierManager.addModifier(c, new EtherealMod());
+                    CardModifierManager.addModifier(c, new GlowTheSoulariumMod());
```
modargo, line 43: I checked it out and this effect looks cool. Good reminder that the cards are Ethereal now (and need to be played this turn if you don't want them to exhaust).

src/main/java/thePackmaster/patches/energyandechopack/OnCreateCardPatch.java
```diff
@@ -0,0 +1,79 @@
...
+import com.megacrit.cardcrawl.cards.AbstractCard;
+import com.megacrit.cardcrawl.cards.blue.DoubleEnergy;
+import com.megacrit.cardcrawl.powers.AbstractPower;
+import thePackmaster.packs.EnergyAndEchoPack;
+import thePackmaster.powers.energyandechopack.DivineTouchPower;
+import thePackmaster.powers.energyandechopack.PanacherPower;
+
+import static thePackmaster.util.Wiz.adp;
+
+public class OnCreateCardPatch {
```
modargo, line 18: This works for most things, but it will miss a few cases because some cards use the effects like `ShowCardAndAddToHandEffect` directly instead of going through one of the `MakeTempCard` actions. For example, the `FlexibleDiscoveryAction` used by `DestinyDraw` uses `ShowCardAndAddToHandEffect`.

To catch everything reliably, I think you can adapt the patch at https://github.com/Jedi515/TheRanger/blob/master/src/TheRanger/patches/TempCardPatches.java, which patches `ShowCardAndAddToDiscardEffect`, `ShowCardAndAddToHandEffect`, and `ShowCardAndAddToDrawPileEffect`. It was written a while back but I still think it's the way to do this (I haven't tried that patch out myself but the idea behind it makes sense to me).

src/main/java/thePackmaster/patches/energyandechopack/TurnStartCheckPatch.java
```diff
@@ -0,0 +1,27 @@
+package thePackmaster.patches.energyandechopack;
+
+import com.evacipated.cardcrawl.modthespire.lib.*;
+import com.megacrit.cardcrawl.actions.GameActionManager;
+import com.megacrit.cardcrawl.characters.AbstractPlayer;
+import javassist.CtBehavior;
+import thePackmaster.packs.EnergyAndEchoPack;
+
+
+@SpirePatch(
```
modargo, line 10: This patch is fine, but FYI it's a bit simpler to make a prefix patch for `clz = AbstractPlayer.class, method = "applyStartOfTurnRelics"`. Avoids the need to define the Locator.

An example of a patch like I'm describing is `DarkLancePatch`.

modargo:
Very cool art! I especially like Voices.

modargo (changes requested):

modargo:
I've taken a first pass through (ignoring Divine Touch and related powers/patches since it's probably getting replaced). Hope all the comments make sense! I'll take another pass whenever things are ready for it.

src/main/java/thePackmaster/cardmodifiers/energyandechopack/EchoedEtherealMod.java
Paul-Sml: this plus i find it better if one effect (echo) adds lines at the same place rather than on each side of the card

src/main/java/thePackmaster/patches/energyandechopack/TurnStartCheckPatch.java
Paul-Sml: Unless an effect calls this function 🤔 
Exactly what I would have done in my Relic pack if it was good enough for me to pursue....
I could change it but I don't really see a reason why, does it lose so much in optimisation ?

modargo: This was just a FYI, so not need to change it.

src/main/java/thePackmaster/patches/energyandechopack/TempCardPatch.java
```diff
@@ -0,0 +1,46 @@
...
+                        AbstractPower pow2 = adp().getPower(PanacherPower.POWER_ID);
+                        if (pow2 != null) {
+                            pow2.onSpecificTrigger();
+                        }
+                        isDone = true;
+                    }
+                });
+
+
+                EnergyAndEchoPack.cardsCreatedThisTurn++;
```
modargo, line 42: Since nothing uses this any more, should delete it.

src/main/resources/anniv5Resources/localization/eng/energyandechopack/Cardstrings.json
```diff
@@ -41,5 +37,10 @@
     "NAME": "Slowing",
     "DESCRIPTION": "At the start of your turn, gain [E] . NL Add !M! *Wounds into your hand.",
     "UPGRADE_DESCRIPTION": "At the start of your turn, gain [E] . NL Add a *Wound into your hand."
+  },
+  "${ModID}:Crystallize": {
+    "NAME": "Crystallize",
+    "DESCRIPTION": "Gain !M! Block for each [E] available.",
+    "EXTENDED_DESCRIPTION": ["Gain !M! Block for each [E] available. NL (Gain !B! Block.)"]
```
modargo, line 44: Small wording change here for base game consistency with cards like Stack.

```suggestion
    "EXTENDED_DESCRIPTION": ["Gain !M! Block for each [E] available. NL (Gains !B! Block.)"]
```

src/main/java/thePackmaster/cards/energyandechopack/Crystallize.java
```diff
@@ -0,0 +1,54 @@
...
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        this.applyPowers();
+        this.addToBot(new GainBlockAction(p, p, this.block));
+    }
+
+    @Override
+    public void applyPowers() {
+        this.baseBlock = EnergyPanel.totalCount * this.magicNumber;
+        super.applyPowers();
+        this.rawDescription = cardStrings.DESCRIPTION + cardStrings.EXTENDED_DESCRIPTION[0];
```
modargo, line 40: With how you have the strings set up, you only need the extended description here (what you currently have will duplicate the description).
```suggestion
        this.rawDescription = cardStrings.EXTENDED_DESCRIPTION[0];
```

src/main/resources/anniv5Resources/localization/eng/energyandechopack/Cardstrings.json
modargo: Though now that I've posted this, I see that Spirit Shield does it the way you have it here. Base game inconsistency... I like "Gains" better than "Gain" but you can argue for either one.

modargo:
Looks good other than the small issues I noted. Once those are fixed I'll see if Gk wants to take a look and then merge.

PR #156, opened 2023-01-15
Fixed Rummage Bug

FieryChocobo:
Upgraded Rummage would only discount 1 card if only 1 card was the highest cost; Now always discounts 2 still starting with the most expensive ones.
If the discount ties a card for new highest cost it can discounted twice.


modargo:
Looks good with the most recent change, thanks for contributing this fix!

PR #159, opened 2023-01-15
H A T S

DarkVexon:
Adds hats. Unlock a Hat for a corresponding Pack by winning with that Pack in your arsenal.

- Hat selection menu in Main Menu UI
- Start with a hat.
- Hats have flavor text.

ALSO:
- Ending comic
- Bellord cards buffed because most people thought they weren't worth it

Notes:
- There are probably a few scaling issues with the hat selection UI, might be worth fixing eventually.
- Fixes major issue (lack of hats)


src/main/java/thePackmaster/hats/HatMenu.java
```diff
@@ -0,0 +1,162 @@
...
+    private static final float BG_X = 525f * Settings.xScale;
+    private static final float BG_Y = Settings.HEIGHT - 40f * Settings.yScale - MENU_BG.getRegionHeight() * BG_Y_SCALE;
+    private static final float DROPDOWN_X = 550f * Settings.xScale;
+    private static final float DROPDOWN_Y = Settings.HEIGHT - 160f * Settings.yScale;
+    private static final float PREVIEW_X = BG_X + (210 * Settings.scale);
+    private static final float PREVIEW_Y = BG_Y + (225 * Settings.scale);
+
+    public static AbstractPlayer dummy;
+
+    private static final boolean UNLOCK_ALL_HATS = true;
```
modargo: Just leaving a note here to set this to false before merging... and to make sure we test with this set to false 😄.

modargo:
Going to ask this here so that both the contributors to this PR and the potential mergers (aka Gk) can see -- do we want to merge this with a partial set of hats (with the expectation of future PRs with more hats), or wait until this PR has hats for all current packs? I see we're around 50% already which is really impressive!

I'm completely fine with merging before we have all the hats, but I want to make sure we're on the same page and that we clearly communicate when there's a desire to merge this PR (vs. whether it's still in progress and we should wait). Obviously with how big this is it makes sense to wait for Gk to be up to take a look too (I've looked through things but need to go to bed at some point), which is why it's helpful to have information recorded on this PR.

src/main/java/thePackmaster/patches/OpeningRunScreenPatch.java
```diff
@@ -18,9 +19,17 @@ public class OpeningRunScreenPatch {
     )
     public static void SetTheThing(NeowEvent __instance) {
         if (AbstractDungeon.player.chosenClass == ThePackmaster.Enums.THE_PACKMASTER) {
-            BaseMod.logger.info("Packmaster is Ready To Do Thing");
-            SpireAnniversary5Mod.openedStarterScreen = false;
-            SpireAnniversary5Mod.doPackSetup = true;
+            if (SpireAnniversary5Mod.allPacksMode) {
+                BaseMod.logger.info("Vex's All Packs Override Enabled. Skipping intro screen");
+                SpireAnniversary5Mod.currentPoolPacks.clear();
+                SpireAnniversary5Mod.currentPoolPacks.addAll(SpireAnniversary5Mod.unfilteredAllPacks);
```
modargo, line 25: Should we respect filtering out packs when setting this up? My immediate reaction is to say that we should, so that "all packs mode" means "all packs except ones that have been filtered out". If someone has a pack disabled for normal play, they probably don't want to see it in all pack mode either.

src/main/java/thePackmaster/patches/MainMenuUIPatch.java
```diff
@@ -148,7 +153,7 @@ public static void Postfix(CharacterOption obj, SpriteBatch sb) {
 
             CharSelectInfo c = ReflectionHacks.getPrivate(obj, CharacterOption.class, "charInfo");
 
-            if (c.player.chosenClass.equals(ThePackmaster.Enums.THE_PACKMASTER) && obj.selected) {
+            if (c != null && c.player.chosenClass.equals(ThePackmaster.Enums.THE_PACKMASTER) && obj.selected) {
```
modargo, line 156: I see this is the null check to fix the bug Piña Colada reported.

README.md
```diff
@@ -77,3 +77,19 @@ Here is a TL;DR:
...
 And finally, as the person merging all PRs other reviewers or I (Gk/erasels) may ask you to make changes to your Pack in case some cards lie outside the power curve or skirt the rules too much.
 Please don't be discouraged if your Pack isn't instantly merged.
+
+### HATS
+
+Packmaster can equip different cosmetic hats, one for every pack. Each one is locked until you've won a run with its respective pack in your pool.
+
+This does not add any mandatory work for packs. There are three ways to get a hat added to your pack:
+
+1) Art it yourself, using the new Hat template PSD.  Add a png export of your hat into the resources/hats folder, with the naming convention of <PackID>Hat.png, and it will automatically be detected and used.
```
modargo, line 87: Let's point people towards the specific file name to reduce the chance of confusion.

```suggestion
1) Art it yourself, using the new Hat template PSD (PackmasterHatTemplate.psd, included in this repo).  Add a png export of your hat into the resources/hats folder, with the naming convention of <PackID>Hat.png, and it will automatically be detected and used.
```

src/main/java/thePackmaster/SpireAnniversary5Mod.java
```diff
@@ -743,35 +786,73 @@ public void update() {
...
+        ModLabeledToggleButton allPacksModeBtn = new ModLabeledToggleButton(configStrings.TEXT[3], 350.0f, configPos, Settings.CREAM_COLOR, FontHelper.charDescFont, allPacksMode, settingsPanel, (label) -> {
+
+        }, (button) -> {
+            allPacksMode = button.enabled;
+            saveAllPacksMode();
+        });
+
+        settingsPanel.addUIElement(allPacksModeBtn);
+
+        BaseMod.registerModBadge(badge, configStrings.TEXT[0], configStrings.TEXT[1], configStrings.TEXT[2], settingsPanel);
```
modargo, line 856: Cool that we have a badge and config menu now.

I'd suggest a different badge icon if it's not too much trouble, something like a backpack. Right now this is using the same book icon that the current pack top panel item uses -- might just be a convenient placeholder? It's also a bit misaligned but I assume that's just because this PR has a lot of moving parts and this one hasn't settled into place yet :smile:.

image

src/main/java/thePackmaster/SpireAnniversary5Mod.java
```diff
@@ -743,35 +786,73 @@ public void update() {
...
         }
-        return packIDs;
+        wrapper.currentPoolPacks = packIDs;
+        wrapper.currentHat = Hats.currentHat;
+        return wrapper;
     }
 
     @Override
-    public void onLoad(ArrayList<String> strings) {
+    public void onLoad(SpireAnniversary5SaveWrapper saved) {
```
modargo, line 801: I bet this will break runs in progress for anyone who started a run before this was changed and saves/recompiles/loads with the new code.

Things are still in development so it isn't that big a deal (people can always check out the older code and finish their run on it if they want), but it means that when we merge this we'll want to warn people. And that we wouldn't want to make a change like this after release (unlikely to be an issue since I doubt we'll make big changes like this after release).

Alternatively, the way to avoid this is to make the Hat into a separate CustomSavable -- you can have more than one, and it doesn't need to only be on the main mod class like it is here. For example, see https://github.com/modargo/corruptthespire/blob/master/src/main/java/corruptthespire/savables/SavableCorruption.java -- which also requires a line in `receivePostInitialize` to add the save field, `BaseMod.addSaveField(SavableCorruption.SaveKey, new SavableCorruption());`. Not really that much work and IMO keeps things better organized (instead of having the pack pool and hat lumped together).

Gk can also weigh in with his thoughts on whether it's worth the work to avoid breaking in progress runs.

modargo:
Took a pass through things, made some comments, and loaded things up locally to see how it actually looked (it loads up and appears to work which is a good start :smile:), but this is where I sign off for the night.

I was planning on doing more Packmaster work and playtesting tomorrow anyway, so when I'm up I'll see where things are and coordinate with Gk/the group on what makes sense to help with.

src/main/java/thePackmaster/patches/OpeningRunScreenPatch.java
erasels: Definitely think we should respect the filtering

src/main/java/thePackmaster/SpireAnniversary5Mod.java
erasels: 100% agree with this as well. Shouldn't have used the custom savable on the main mod file to begin with. Would prefer use BaseMod.addCustomSaveField

erasels:
Pretty much agree with everything modargo said.

One small thing is that we have our own logger defined in SpireAnniversary5Mod, please use that instead of the BaseMod logger

erasels:
> Going to ask this here so that both the contributors to this PR and the potential mergers (aka Gk) can see -- do we want to merge this with a partial set of hats (with the expectation of future PRs with more hats), or wait until this PR has hats for all current packs? I see we're around 50% already which is really impressive!
> 
> I'm completely fine with merging before we have all the hats, but I want to make sure we're on the same page and that we clearly communicate when there's a desire to merge this PR (vs. whether it's still in progress and we should wait). Obviously with how big this is it makes sense to wait for Gk to be up to take a look too (I've looked through things but need to go to bed at some point), which is why it's helpful to have information recorded on this PR.

As for this. I think merging with missing hats is fine here as long as this system has fail safes for not crashing when a hat doesn't exist for a pack.
I think merging it now will show people that it exists and might encourage people to add their own hat.

erasels:
Just a note, I would've really preferred if these had been separate pull requests. Having the hat system, all pack run, character tweaks, art stuff and one frame mode all in one PR is terrible for double checking and makes it hard to watch out for errors.

src/main/java/thePackmaster/patches/OpeningRunScreenPatch.java
DarkVexon: And done! Good call

src/main/java/thePackmaster/SpireAnniversary5Mod.java
DarkVexon: Yep, this was a placeholder. I figured someone with actual art skills would set up some art, but I'll actually find an aligned version so it doesn't look this bad haha

src/main/java/thePackmaster/SpireAnniversary5Mod.java
DarkVexon: OK, I figured a save wrapper may not have been the best option. I'll see about fixing this potentially based off the example; never used a CustomSavable outside of mod file contexts before

PR #162, opened 2023-01-15
Last card replacement I promise

bkerns724:


src/main/java/thePackmaster/SpireAnniversary5Mod.java
```diff
@@ -506,8 +518,58 @@ public void receiveOnBattleStart(AbstractRoom room) {
         CLAW_SHARP_TRACKER = 0;
         combatExhausts = 0;
     }
-    
-	@Override
+
+    @Override
+    public void receiveOnPlayerTurnStart() {
```
erasels, line 523: Please put this logic into another class and call it from here.
So for example.
`public void receiveOnPlayerTurnStart() {
Leprachaun.receieveOnPlayerTurnStart();
}`
The main mod file is getting a little crowded

src/main/java/thePackmaster/powers/summonspack/JinxPower.java
```diff
@@ -0,0 +1,37 @@
...
+    public static final String NAME = CardCrawlGame.languagePack.getPowerStrings(POWER_ID).NAME;
+    public static final String[] DESCRIPTIONS = CardCrawlGame.languagePack.getPowerStrings(POWER_ID).DESCRIPTIONS;
+
+    public JinxPower(AbstractCreature owner, int amount) {
+        super(POWER_ID, NAME, AbstractPower.PowerType.DEBUFF, false, owner, amount);
+    }
+
+    @Override
+    public boolean onReceivePower(AbstractPower pow, AbstractCreature target, AbstractCreature source) {
+        if (pow.type != AbstractPower.PowerType.DEBUFF || pow.owner == source || pow.ID.equals(GainStrengthPower.POWER_ID)
```
erasels, line 26:
```suggestion
        if (pow.type == AbstractPower.PowerType.BUFF || pow.owner == source || pow.ID.equals(GainStrengthPower.POWER_ID)
```
Please don't check like that since you could catch neutral power type which would be bad

erasels (changes requested):

PR #165, opened 2023-01-15
Add Thievery pack

Celicath:
This pack has cards with the 'stealing' theme, and a few cards that synergizes with them. Few cards originated from the Dragon Tamer mod, and others are scrapped cards from the Copycat (both with small tweaks).

The cards probably don't make an archetype, and are more like general purpose cards. There are few small synergies though. 

I did the basic testing, but not full runs or intensive tweaks. I made the PR now because the cutoff is soon, and might tweak cards later. Feedbacks are welcome as well.


src/main/java/thePackmaster/cards/thieverypack/AbstractThieveryCard.java
```diff
@@ -0,0 +1,18 @@
...
+		super(ID, cost, type, rarity, target);
+		setBackgroundTexture(
+			"anniv5Resources/images/512/thieverypack/" + type.name().toLowerCase() + ".png",
+			"anniv5Resources/images/1024/thieverypack/" + type.name().toLowerCase() + ".png"
+		);
+	}
+
+	@Override
+	public void upp() {
+		// Yay
```
erasels, line 16: Why do this and then use upgrade instead? This way you could've skipped the !upgraded check and the description/name upgrading

src/main/resources/anniv5Resources/localization/eng/thieverypack/Potionstrings.json
```diff
@@ -0,0 +1,9 @@
+{
+  "${ModID}:DivinePotion": {
+    "NAME": "Divine Potion",
+    "DESCRIPTIONS": [
+      "Deal #b",
```
erasels, line 5: Gain?

src/main/resources/anniv5Resources/localization/eng/thieverypack/UIstrings.json
```diff
@@ -0,0 +1,9 @@
+{
+  "${ModID}:ThieveryPack": {
+    "TEXT": [
+      "Thief, the Deceiver",
+      "No need to prepare all yourself; stealing can get what you want.",
```
erasels, line 5:
```suggestion
      "No need to prepare everything yourself; stealing can get what you want.",
```

src/main/resources/anniv5Resources/localization/eng/thieverypack/Cardstrings.json
```diff
@@ -0,0 +1,46 @@
+{
```
erasels, line 1: Kind of surprised there isn't a gold stealing card in this.

I like the flavor but not that enemies fulfill the conditions so some of these cards are hard to justify

erasels (changes requested):

src/main/java/thePackmaster/cards/thieverypack/AbstractThieveryCard.java
Celicath: Yeah sorry, didn't like the name of `upp`, and since I copied my old code, it's better for me to directly use `upgrade` myself. I test the description change on upgrade so that's fine. I can fix my code to use `upp` if you really want though.

erasels: Nah, just wanted to mention it. All good

erasels (approved):

PR #174, opened 2023-01-15
Packmaster Class Potions (and one general potion)

mikemayhemdev:
Also switched Cheapen art since the former art is now going to be used by orbs pack


src/main/resources/anniv5Resources/localization/eng/Potionstrings.json
```diff
@@ -0,0 +1,34 @@
+{
+  "${ModID}:BoosterBrew": {
+    "NAME": "Booster Brew",
+    "DESCRIPTIONS": [
+      "Choose 1 of 3 random cards from different Packs. Add #b1 copy of it to your hand. It costs 0 this turn.",
```
modargo, line 5: Need highlighting for the numbers.
```suggestion
      "Choose #b1 of #b3 random cards from different Packs. Add #b1 copy of it to your hand. It costs 0 this turn.",
```

src/main/resources/anniv5Resources/localization/eng/Potionstrings.json
```diff
@@ -0,0 +1,34 @@
...
+      " Repeat this effect #b",
+      " times."
+    ]
+  },
+  "${ModID}:ModdersDelight": {
+    "NAME": "Modder's Delight",
+    "DESCRIPTIONS": [
+      "Return all cards in your discard pile to your hand. Draw #b",
```
modargo, line 13: The text and the effect for this don't line up -- looking at the code, it shuffles the discard pile back into your draw pile, then draws. Is that the intended effect? What I see the code doing makes more sense to me than what's written here.

If so this can use the wording of Deep Breath combined with the wording Booster Tutor: `Shuffle your discard pile into your draw pile. NL Put #bX random cards from different Packs from your draw pile into your hand.`

src/main/java/thePackmaster/potions/SmithingOil.java
```diff
@@ -0,0 +1,74 @@
...
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.helpers.PowerTip;
+import com.megacrit.cardcrawl.localization.PotionStrings;
+import thePackmaster.SpireAnniversary5Mod;
+import thePackmaster.util.Wiz;
+
+import static thePackmaster.util.Wiz.atb;
+
+
+public class SmithingOil extends CustomPotion {
```
modargo, line 20: This is an extremely powerful effect. I know I'd be very excited to see an upgrade drop from a fight, which is essentially what this is. But it's so powerful that I wonder if it's too much to include. Thoughts?

If we do keep this, since this potion has a permanent benefit, this will need a patch for `AbstractDungeon.returnRandomPotion`. We need the same check as for Fruit Juice (not allowing it when `limited` is true), so that this potion can't be randomly generated by Alchemize or any other effect like that.

src/main/java/thePackmaster/potions/PackInAJar.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        addToBot(new EasyModalChoiceAction(packCards, 1, CardCrawlGame.languagePack.getUIString(makeID("ModalChoice")).TEXT[0]));
+    }
+
+    public void action(AbstractCardPack pack) {
+        addToBot(new DiscardAction(AbstractDungeon.player, AbstractDungeon.player, Wiz.hand().size(), true, false));
+        for (int i = 0; i < potency; i++) {
+            for (AbstractCard c : pack.cards) {
+                if (c.rarity == AbstractCard.CardRarity.COMMON || c.rarity == AbstractCard.CardRarity.UNCOMMON || c.rarity == AbstractCard.CardRarity.RARE) {
```
modargo, line 66: We should also filter out cards that have the `CardTags.HEALING` tag, since those should never be generated in combat.

src/main/resources/anniv5Resources/localization/eng/Potionstrings.json
```diff
@@ -0,0 +1,34 @@
...
+      " cards from different Packs in your draw pile."
+    ]
+  },
+  "${ModID}:PackInAJar": {
+    "NAME": "Pack in a Jar",
+    "DESCRIPTIONS": [
+      "Discard your hand, then choose 1 of 3 Packs. Add #b",
```
modargo, line 20:
```suggestion
      "Discard your hand, then choose #b1 of #b3 Packs. Add #b",
```

src/main/java/thePackmaster/potions/PackInAJar.java
```diff
@@ -0,0 +1,82 @@
...
+        for (AbstractCardPack pack : choices) {
+            packCards.add(new PackmasterModalChoiceCard(pack.previewPackCard.cardID, pack.previewPackCard.name, pack.previewPackCard.rawDescription, true, () -> action(pack)));
+        }
+
+        addToBot(new EasyModalChoiceAction(packCards, 1, CardCrawlGame.languagePack.getUIString(makeID("ModalChoice")).TEXT[0]));
+    }
+
+    public void action(AbstractCardPack pack) {
+        addToBot(new DiscardAction(AbstractDungeon.player, AbstractDungeon.player, Wiz.hand().size(), true, false));
+        for (int i = 0; i < potency; i++) {
```
modargo, line 64: The potency on this is pretty pointless almost all of the time, right? A pack is going to fill or mostly fill your hand, unless you have a max hand size increasing effect.

There are potions like Gambler's Brew that just aren't affected by potency increases -- maybe Pack Rip should be like that too.

src/main/java/thePackmaster/potions/PackInAJar.java
```diff
@@ -0,0 +1,82 @@
...
+
+
+public class PackInAJar extends CustomPotion {
+    public static final String POTION_ID = SpireAnniversary5Mod.makeID("PackInAJar");
+    private static final PotionStrings potionStrings = CardCrawlGame.languagePack.getPotionString(POTION_ID);
+    public static final String NAME = potionStrings.NAME;
+    public static final String[] DESCRIPTIONS = potionStrings.DESCRIPTIONS;
+
+    public PackInAJar() {
+        super(NAME, POTION_ID, PotionRarity.UNCOMMON, PotionSize.GHOST, PotionColor.NONE);
```
modargo, line 32: This could be rare, and in my opinion it's powerful enough that it should be. Getting a whole hand of cards is a really unique effect. I'm happy to have this potion exist as a very powerful one-time-use resource that will often make a decisive difference even in the hardest fights -- but if I'm assessing the impact this will have correctly, it fits better with other rare potions.

And assuming we make Booster Brew a Packmaster-specific potion, I think that fits better as Packmaster's uncommon potion, leaving Pack Rip and Smithing Oil to fight it out for the rare slot. Not saying we need to cut one, though as I mentioned in another comment I think we should discuss whether we want to keep Smithing Oil (I'm fine if we do).

src/main/java/thePackmaster/SpireAnniversary5Mod.java
```diff
@@ -409,6 +418,10 @@ public static void addPotions() {
             whitelist.accept(DrawClawsPotion.POTION_ID);
             whitelist.accept(GenerateClawsPotion.POTION_ID);
             whitelist.accept(DivinePotion.POTION_ID);
+            whitelist.accept(BoosterBrew.POTION_ID);
+            whitelist.accept(ModdersDelight.POTION_ID);
+            whitelist.accept(PackInAJar.POTION_ID);
+            whitelist.accept(SmithingOil.POTION_ID);
```
modargo, line 424: Glad I put the wide potion stuff in place.

src/main/java/thePackmaster/potions/ModdersDelight.java
```diff
@@ -0,0 +1,56 @@
...
+
+        addToBot(new BoosterTutorAction(potency));
+    }
+
+    public CustomPotion makeCopy() {
+        return new ModdersDelight();
+    }
+
+    public int getPotency(int ascensionLevel) {
+        return 4;
```
modargo, line 52: I think this would still be strong at 3. Similar to draw potion, and the shuffle effect is hard to assess -- sometimes it'll be good, sometimes it won't.

modargo:
Glad we remembered these, a character isn't really complete without character-specific potions. I took a first pass through with the goal of engaging on both the code and on balance/functionality, since this is the first time we've really discussed potion ideas since the start of the project. I'm certainly not the final word on balance but as you can probably tell it's something I'm always interested in discussing, and that's the goal of my comments about it :smile:.

This is one where I want to let Gk take a look over it as well once he's up and able to engage, to get another opinion on any balance/functionality questions if nothing else.

src/main/java/thePackmaster/potions/SmithingOil.java
mikemayhemdev: Yeah, the comparison to Fruit Juice is what I was going for here.  I'm open to other ideas.  The Smithing Hammer gave an 'in' for theming to Packmaster, and I wanted the Rare to be as generally useful as possible.

src/main/java/thePackmaster/potions/PackInAJar.java
mikemayhemdev: Hm... fundamentally I would agree, but this makes the theming of the potion invalid.  It's the revival of Pack Rip, which would feel odd if it sometimes can give 10 cards and sometimes 8 because of HEALING.  I would say this is a case where it should be an exception to the rule.

modargo: I understand your point, which means we have a dilemma. Happily, I have a suggestion that can square this circle.

To accept that this potion can give healing tagged cards, we just need for the potion itself to be given the same treatment as Fruit Juice and Smithing Oil -- no creating it from in-combat potion generation. That will resolve any issues I have with this, because it will prevent the thing that can provide permanent benefit (in this case, the Pack in a Jar potion) from being generated during combat.

mikemayhemdev: Good call, I have made it so!

src/main/resources/anniv5Resources/localization/eng/Potionstrings.json
```diff
@@ -0,0 +1,34 @@
+{
+  "${ModID}:BoosterBrew": {
+    "NAME": "Booster Brew",
+    "DESCRIPTIONS": [
+      "Choose #b1 of #b3 random cards from different Packs. Add #b1 copy of it to your hand. It costs 0 this turn.",
```
modargo, line 5:
```suggestion
      "Choose #b1 of #b3 random cards from different Packs. Add #b1 copy of it to your hand. It costs #b0 this turn.",
```

src/main/resources/anniv5Resources/localization/eng/UIstrings.json
```diff
@@ -105,7 +105,8 @@
       "StS Modding Community",
       "Adds a new character: The Packmaster.",
       "ALL PACKS MODE. Ignores Custom Draft",
-      "One Frame For All. All Pack cards have normal Packmaster frames (Requires Restart)."
+      "One Frame For All. All Pack cards have normal Packmaster frames (Requires Restart).",
+      "Share Content. Some Relics/ Potions can appear on other characters (Requires Restart)."
```
modargo, line 109: Happy with where we landed with this -- people can opt in if they want. Also good to see we call out that it requires a restart.

src/main/resources/anniv5Resources/localization/eng/madsciencepack/Cardstrings.json
```diff
@@ -1,45 +1,45 @@
 {
   "${ModID}:Sharpen": {
     "NAME": "Honing Strike",
-    "DESCRIPTION": "Deal !D! damage. NL ${ModID}:Modify a random Attack in hand: 'Add !M! Damage'"
+    "DESCRIPTION": "Deal !D! damage. NL ${ModID}:Modify a random Attack in hand: \"Add !M! Damage\""
```
modargo, line 4: I see that this is the solution to the odd text/keyword parsing issues you were seeing. Glad you got that figured out.

modargo:
Things look good in terms of the code.

Given the discussion in Discord I'll wait to merge until the Pack in a Jar removal/Booster Brew getting a new effect. I can take a look at that whenever it's ready, but I'll probably let Gk take the final pass on this (if he wants to) and be the one to press merge, unless you really want to merge it in tonight.

mikemayhemdev:
@modargo That update is now complete.  No need to rush a merge in if you're not comfortable, but it was a pretty easy one to change given Chrysalis is not a custom action.

modargo:
> @modargo That update is now complete. No need to rush a merge in if you're not comfortable, but it was a pretty easy one to change given Chrysalis is not a custom action.

Since most other things have been complex I forgot how easy it was to do what Chrysalis does :smile:.

src/main/java/thePackmaster/potions/BoosterBrew.java
```diff
@@ -0,0 +1,83 @@
...
+            packCards.add(new PackmasterModalChoiceCard(pack.previewPackCard.cardID, pack.previewPackCard.name, pack.previewPackCard.rawDescription, true, () -> action(pack)));
+        }
+
+        addToBot(new EasyModalChoiceAction(packCards, 1, CardCrawlGame.languagePack.getUIString(makeID("ModalChoice")).TEXT[0]));
+    }
+
+
+    public void action(AbstractCardPack pack) {
+        for (int i = 0; i < potency; i++) {
+            AbstractCard card = SpireAnniversary5Mod.getRandomCardFromPack(pack);
```
modargo, line 62: I was going to ask about duplicates, then I realized that Chrysalis can create multiple copies of the same card, so it's fine that this can too.

PR #176, opened 2023-01-16
Biting Cold & Into the Breach Hats

ActuallyEnbeon:
Adds Hats for the Biting Cold and Into the Breach packs
Also a few other commits that I guess got duplicated since their changes are already in main lol


modargo:
Don't worry about the commit duplication, it happened because we squash merged a few PRs. Git is still intelligent enough to recognize that they represent changes that were already incorporated into the main repo.

PR #197, opened 2023-01-17
Mostly vfx/sfx tweaks, some small balance adjustments

bkerns724:
Color damage numbers now fade to white
Panda passive slight bump
Fixed leprechaun+ cardstring
Training now evokes one more time but consumes orb
Messed with some evoke timing on Training
Bee attacks no longer animate on dead enemies
Swarm of Bees audio fades out faster
Shorter elephant animation when sound off


src/main/java/thePackmaster/cards/summonspack/Leprechaun.java
```diff
@@ -60,5 +60,6 @@ public List<TooltipInfo> getCustomTooltips() {
     @Override
     public void upp() {
         upMagic(UPGRADE_MAGIC);
+        uDesc();
```
modargo, line 63: Updating the description should be handled by `AbstractPackmasterCard` as long as an upgrade description is defined, so I don't think you need this call.

PR #200, opened 2023-01-17
-more hats, batter tweaks

AlexMdle:


modargo:
Changes looks good and the new hats look great! Took a look over things and going to merge this in.

PR #205, opened 2023-01-17
rework crystal resonance

JohnnyDevo:


src/main/java/thePackmaster/powers/transmutationpack/CrystalResonancePower.java
```diff
@@ -20,11 +24,11 @@ public CrystalResonancePower(AbstractCreature owner, int amount) {
...
     @Override
     public void updateDescription() {
-        this.description = amount == 1 ? DESCRIPTIONS[0] : DESCRIPTIONS[1] + amount + DESCRIPTIONS[2];
+        description = DESCRIPTIONS[0] + amount + DESCRIPTIONS[1];
     }
 
     @Override
-    public void affectTransmutedCard(AbstractCard newCard) {
-        addToBot(new DrawCardAction(amount));
+    public void onTransmute(TransmuteCardAction action) {
```
modargo, line 31: I tested this out locally -- is it intended that if you play a card like Glacial Synthesis+, but choose to transmute no cards, you still get the strength?

I looked at the code and understand why it works this way, but just reading the cards I would expect on transmute effects to only trigger when cards were actually transmuted. I also note that Glacial Synthesis still plays a sound effects and waits on the screen like there are cards being transmuted, even when no cards are being transmuted. So my assumption is that this is a bug stemming from the slightly unusual scenario and that it should be fixed.

The call to the `onTransmute` should probably be moved to the place in the action where a non-zero number of cards are transmuted.

modargo (changes requested):

PR #215, opened 2023-01-17
Thieverypack(Thief, the Deceiver) Rework

Celicath:
I did a few runs with my pack. My thoughts:
- Strength stealing feels great. There are many enemies that gain Strength themselves, and it really feels like you are stealing their powers.
- Magnet and Null Hammer feels really bad because they miss a lot.
- Cunning power is OP
- Other cards are just ok, but they could benefit from small number buffs.

With this in mind, I made these changes:
| Old | New | Notes |
|-------|-------|------|
| image | image | Cunning Poison is a completely different, gold-stealing card. Now the thief can steal money. |
| image | image | Magnet felt good when it actually stole Artifact, but there are only a few enemies that have them. Now magnet can steal all metal-ish powers, hopefully giving it more chances. Magnet upgrades to ALL enemies. |
| image | image | This was almost a pure damage card, now it's more spicy. |
| image image | image image | They needed this buff. |
| image | image | Thievery Mastery no longer stacks. Boo! But this allowed me to create gold-stealing card, because it's much harder to copy Skills than Powers. You can still get infinite gold by copying Cunning Poison over and over, it's unfortunate but I didn't want to make the card super complex to prevent this. |

Cardbacks are slightly adjusted, a few bugs were fixed.


erasels:
Magnet -> from enemy = from the enemy
Cunning poison -> highest HP
Thanks for the great explanation, I appreciate the effort you put into this.

src/main/java/thePackmaster/actions/thieverypack/StealPowerAction.java
```diff
@@ -181,63 +196,104 @@ public void update() {
...
+				return new RitualPower(c, gainAmount, true);
+			case AngerPower.POWER_ID:
+				return new AngerPower(c, gainAmount);
+			case StasisPower.POWER_ID:
+				AbstractCard card = ReflectionHacks.getPrivate(pow, StasisPower.class, "card");
+				if (card != null) {
+					return new StasisPower(c, card) {
+						@Override
+						public void onInitialApplication() {
+							if (AbstractDungeon.player.hand.size() != 10) {
```
erasels, line 284:
```suggestion
							if (AbstractDungeon.player.hand.size() != BaseMod.MAX_HAND_SIZE) {
```
requires import

erasels (changes requested):

Celicath:
Thanks for the suggestions, I changed them.

src/main/java/thePackmaster/actions/thieverypack/StealPowerAction.java
```diff
@@ -281,7 +282,7 @@ private AbstractPower getDuplicatedPower(AbstractCreature c, AbstractPower pow,
 					return new StasisPower(c, card) {
 						@Override
 						public void onInitialApplication() {
-							if (AbstractDungeon.player.hand.size() != 10) {
+							if (AbstractDungeon.player.hand.size() != BaseMod.MAX_HAND_SIZE) {
```
erasels, line 285:
```suggestion
							if (AbstractDungeon.player.hand.size() < BaseMod.MAX_HAND_SIZE) {
```

erasels (approved):

PR #218, opened 2023-01-17
Pinnacle Pack by SandTag & small fixes for Anomaly pack

SandTag:
A full pack and previously agreed design changes.

Two small fixes for the Anomaly pack magic number issue.


src/main/java/thePackmaster/cards/pinnaclepack/MarvelousFeastUncommonPower.java
```diff
@@ -0,0 +1,37 @@
...
+import thePackmaster.powers.pinnaclepack.Cooking_;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class MarvelousFeastUncommonPower extends AbstractPinnacleCard {
+
+    public final static String ID = makeID("MarvelousFeastUncommonPower");
+    private boolean fixUpgradeCheck = false;
+
+    public MarvelousFeastUncommonPower() {
```
modargo, line 17: *Marvelous Feast*
The design of this card has multiple issues.

This card uses multiple keywords when it should instead write out its effects -- it should be changed to clearly and directly describe what it does.. This card should also should preview the Croquette cards using MultiCardPreview (e.g. put `MultiCardPreview.add(this, new FishyCroquettesSpecialColourless(), new FriendshipCroquettesSpecialColourless(), etc);` in the constructor, with the croquettes ordered by the order they show up).

This means that the card will read something like: "At the end of the next four turns, add a *Croquette to your hand, in the order *Friendship, *Fishy, *Meaty, *Mystery." None of those need to be keywords, since the preview cards will show the croqeuttes.

Finally, the upgrade to this card goes against how upgrades normally work (by increasing the cost of the card; this is another thing that might be okay in a pack where it was the only odd thing, but in a pack with many odd things it's an issue). Instead of increasing cost, the upgrade should keep the same cost, and the upgraded Croquettes should be rebalanced as needed to fit that by having the upgrades increase numbers by less than the currently do.

src/main/java/thePackmaster/cards/pinnaclepack/PocketTournamentUncommonSkill.java
```diff
@@ -0,0 +1,29 @@
...
+import thePackmaster.cards.pinnaclepack.actions.MinigameAction;
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class PocketTournamentUncommonSkill extends AbstractPinnacleCard {
+
+    public final static String ID = makeID("PocketTournamentUncommonSkill");
+    private static final int MAGIC = 0;
+    private static final int UPGRADE_MAGIC = 2;
+
+    public PocketTournamentUncommonSkill() {
```
modargo, line 14: *Pocket Tournament*
Instead of the using a keyword, the effect of this card should be written out on the card.

This means that the cards text will read: "Randomly, 2X times: NL Draw 1 card next turn. NL Gain [E] next turn. NL Gain 3 Block next turn. NL Gain 5 Vigor. NL Exhaust." This is at the limit of 6 lines per card, and it's not great that even after these suggestions there will be multiple long/complex cards like this still around, but it will help.

src/main/java/thePackmaster/cards/pinnaclepack/Rea__dyCommonSkill.java
```diff
@@ -0,0 +1,40 @@
...
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class Rea__dyCommonSkill extends AbstractPinnacleCard {
+
+    public final static String ID = makeID("Rea__dyCommonSkill");
+
+    private static final int MAGIC = 3;
+    private static final int MAGIC2 = 2;
+
+    public Rea__dyCommonSkill() {
```
modargo, line 20: *Re__y*
The upgrade on this adds a new effect, which is unusual and makes it harder to players to evaluate the card when they see it. (Again, you can find other examples of this, but the issue with this pack is that there are so many unusual and complex things.) Also, the upgrade both changes Exhaust to Exhaustive and adds this new effect, which is overkill.

The cleanest way to simplify this is to change the upgrade so that it no longer adds Draw 1 card. Instead, it can increase the card's other effects.

src/main/java/thePackmaster/cards/pinnaclepack/StoneSwordCommonAttack.java
```diff
@@ -0,0 +1,39 @@
...
+import com.megacrit.cardcrawl.actions.common.ApplyPowerAction;
+import com.megacrit.cardcrawl.actions.common.DamageAction;
+import com.megacrit.cardcrawl.cards.DamageInfo;
+import com.megacrit.cardcrawl.characters.AbstractPlayer;
+import com.megacrit.cardcrawl.monsters.AbstractMonster;
+import com.megacrit.cardcrawl.powers.DexterityPower;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class StoneSwordCommonAttack extends AbstractPinnacleCard {
```
modargo, line 13: *Stone Sword*
The overall design of this card violates too many design principles for it to be in the pack. As it stands, this card will typically have one of two states -- unusable (in all the situations where you can't play it because its a 3-cost card with minimal immediate impact) or game-breaking (when you have Snecko Eye or other cost reduction that lets you abuse the non-exhausting common-rarity dexterity increase on an attack to rapidly outscale enemies).

The best course of action here is to remove this card and replace it with a new card with a new design. I recommend that you create something that is well within the guidelines for complexity.

src/main/java/thePackmaster/cards/pinnaclepack/TheSuckamidRareSkill.java
```diff
@@ -0,0 +1,50 @@
...
+import com.megacrit.cardcrawl.actions.common.HealAction;
+import com.megacrit.cardcrawl.cards.AbstractCard;
+import com.megacrit.cardcrawl.characters.AbstractPlayer;
+import com.megacrit.cardcrawl.monsters.AbstractMonster;
+import com.megacrit.cardcrawl.powers.EquilibriumPower;
+import thePackmaster.powers.pinnaclepack.Capacitor;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class TheSuckamidRareSkill extends AbstractPinnacleCard {
```
modargo, line 14: *The Suckamid*
The overall design of this card has two issues: Purge and Capacitor.

The use of Purge is a problem because Purge is a mechanic that is not used on normal cards by the base game. It also has several unintuitive/buggy interactions with other effects. Instead of using Purge, this card should Exhaust. While there may be ways to recover this card from Exhaust that will be abusable with the right packs, it is still preferable to use Exhaust instead of Purge here.

Instead of the Capacitor keyword, which shares its name with a base game power and isn't necessary to keyword anyway, the text for the effect should be written out. Doing this will also make it clear that the text for this card is close to or at the text length limits, which is part of the overall issue with this pack. If the text length is over the limit, the increase to maximum hand size is what I'd suggest removing.

src/main/java/thePackmaster/cards/pinnaclepack/TopOfTheSkyUncommonAttack.java
```diff
@@ -0,0 +1,47 @@
...
+import com.megacrit.cardcrawl.actions.animations.VFXAction;
+import com.megacrit.cardcrawl.actions.common.DamageAction;
+import com.megacrit.cardcrawl.actions.common.GainBlockAction;
+import com.megacrit.cardcrawl.cards.DamageInfo;
+import com.megacrit.cardcrawl.characters.AbstractPlayer;
+import com.megacrit.cardcrawl.monsters.AbstractMonster;
+import com.megacrit.cardcrawl.vfx.combat.IronWaveEffect;
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class TopOfTheSkyUncommonAttack extends AbstractPinnacleCard {
```
modargo, line 13: *Top of the Sky*
The overall design of this card is okay (though it's another card that runs right up to the text length limits), but the upgrade is out of line with how good it is -- considering what you get if you play the card until it exhausts, it goes from 4 damage/2 block to 9 damage/6 block. This is due to the design of returning to your hand when you play it, which chanages how we have to evaluate the card.

The cleanest ways to simplify this are to either (a) remove the increase to exhaustive or (b) remove the upgrade to damage/block.

src/main/java/thePackmaster/cards/pinnaclepack/MeatyCroquettesSpecialColourless.java
```diff
@@ -0,0 +1,31 @@
...
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class MeatyCroquettesSpecialColourless extends AbstractPinnacleCard {
+
+    public final static String ID = makeID("MeatyCroquettesSpecialColourless");
+    private static final int MAGIC = 3;
+    private static final int UPGRADE_MAGIC = 3;
+
+
+    public MeatyCroquettesSpecialColourless() {
```
modargo, line 17: *Meaty Croquette*
This is fine for a token card, but it shouldn't be a power that exhausts (powers don't exhaust).

Making this a skill like the other Croquettes would increase the similarity between the Croquettes, which helps reduce their complexity, and would fix this issue.

src/main/java/thePackmaster/cards/pinnaclepack/FishyCroquettesSpecialColourless.java
```diff
@@ -0,0 +1,31 @@
...
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class FishyCroquettesSpecialColourless extends AbstractPinnacleCard {
+
+    public final static String ID = makeID("FishyCroquettesSpecialColourless");
+    private static final int MAGIC = 2;
+    private static final int UPGRADE_MAGIC = 2;
+
+
+    public FishyCroquettesSpecialColourless() {
```
modargo, line 16: *Fishy Croquette*
This is fine for a token card.

src/main/java/thePackmaster/cards/pinnaclepack/FriendshipCroquettesSpecialColourless.java
```diff
@@ -0,0 +1,39 @@
...
+public class FriendshipCroquettesSpecialColourless extends AbstractPinnacleCard {
+
+    public final static String ID = makeID("FriendshipCroquettesSpecialColourless");
+    private static final int MAGIC = 2;
+    private static final int UPGRADE_MAGIC = 2;
+    private static final int BLOCK = 6;
+    private static final int UPGRADE_BLOCK = 6;
+
+
+    public FriendshipCroquettesSpecialColourless() {
```
modargo, line 18: *Friendship Croquette*
This is too complex for a token card and too different from the other Croquette cards. It should be simplified. The most straightforward way to do this is by removing the block increase on retain (which is something none of the other Croquettes have).

src/main/java/thePackmaster/cards/pinnaclepack/MysteryCroquettesSpecialColourless.java
```diff
@@ -0,0 +1,30 @@
...
+import com.megacrit.cardcrawl.powers.VulnerablePower;
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class MysteryCroquettesSpecialColourless extends AbstractPinnacleCard {
+
+    public final static String ID = makeID("MysteryCroquettesSpecialColourless");
+    private static final int MAGIC = 1;
+    private static final int UPGRADE_MAGIC = 99;
+
+    public MysteryCroquettesSpecialColourless() {
```
modargo, line 15: *Mystery Croquette*
This is fine for a token card but the upgrade is out of line and needs to change. I'd recommend making this same values as the one that gives weak.

src/main/java/thePackmaster/cards/anomalypack/SouleaterStrike.java
```diff
@@ -14,10 +14,11 @@ public class SouleaterStrike extends AbstractAnomalyCard {
     public static final String ID = SpireAnniversary5Mod.makeID("SouleaterStrike");
     private static final int COST = 1;
     private static final int ATTACK_DMG = 13;
+    private static final int MAGIC = 13;
 
     public SouleaterStrike() {
```
modargo, line 19: The creator of the Anomaly pack made their own fix for this and ThoughtWeaving, so you'll need to resolve the merge conflicts here.

modargo (changes requested):
Hi SandTag, thanks for putting together this PR for the Pinnacle pack.

I've taken a look through things, and this pack will need some changes to the cards in it before it can be merged.

The issue with this pack is that it's at (or really, over) the complexity limit, and needs to be simplified. This is due to a combination of factors:
* Several keywords that are used on only one card and have enough text that the card would be 6 or more lines if the keywords were written out
* Several cards that diverge significantly from typical Slay the Spare card design and balance principles (e.g. unusual upgrades, cards that are either useless or broken with very little middle ground)
* A card that uses 4 token cards (which we grandfathered in despite putting a 3-token limit in place when it became clear that many tokens was becoming a complexity problem)
* Many cards are at or close to the individual card complexity limit due to have many separate effects and/or long text length

If only one or two of these factors were present, the pack might be within the guidelines from a design perspective. After all, the guidelines are written to have some flexiblity, and if a pack is in good shape overall, having a few complex areas is fine. But with a case like this, where the boundaries are being pushed in many different ways, the result is a pack that needs to be simplified before it can be merged.

I'm sure you can point to other packs with examples of cards like some of yours, and mods other than the Packmaster that inspired these cards, but that doesn't change the complexity level of the Pinnacle pack or the guidelines for the Packmaster character. Let's keep this discussion focused on this pack and what we can do with it.

To help with this process, I have gone through and commented on each card with suggestions for changes to address the issues, ranging from wording changes to reworks. These suggestions tie back to the factors I mention above -- removing and writing out unnecessary keywords, bringing design and balance in line with normal principles, etc.

To be clear, taking the suggestions I've made isn't the only way to get this into a state that can be merged, and there's certainly room for discussion. The suggestions are intended to provide a clear path to achieve the goal of getting this merged -- one that I spent a decent amount of time thinking about and discussed with Gk as well. They're also a map to where the over-the-line complexity of this pack is coming from -- meaning that some kind of solution needs to be found for the areas highlighted.

In addition, I've left feedback on a few other areas -- code, wording, and frames. Most of it is more straightforward, but that feedback will also need to be addressed prior to merging.

Given the volume of feedback here, it may be challenging to have this PR ready for the initial release. We'll do our best to take a second pass and get this merged in when you're ready with changes, whether that's before or after the deadline for the initial release.

Please let us know if you have any questions in whichever of Github or Discord you're most comfortable with.

FYI Gk @erasels.

src/main/java/thePackmaster/cards/pinnaclepack/FriendshipCroquettesSpecialColourless.java
SandTag: No, i agreed with GK the croquettes were fine in advance.

PR #225, opened 2023-01-18
New Pack: Content Creators! 

DarkVexon:
By MichaelMayhem and I. Adds 10 cards themed around modded Spire content creators.

Also has a pack image for the Oddities pack and a hat because I forgot those and maybe some visual effects. 

Also it saves the Hat you had equipped between save and load. Fixed Heart wins not refreshing the Hat dropdown, and a few issues with Locked hats!


src/main/java/thePackmaster/patches/contentcreatorpack/CharacterTwoAmountPowerStackPatch.java
```diff
@@ -0,0 +1,27 @@
...
+import javassist.CtBehavior;
+import thePackmaster.powers.contentcreatorpack.FrostprimePower;
+
+@SpirePatch(clz = ApplyPowerAction.class, method = "update")
+public class CharacterTwoAmountPowerStackPatch {
+    @SpireInsertPatch(locator = Locator.class, localvars = { "p" })
+    public static void stackAmount2(ApplyPowerAction __instance, AbstractPower p) {
+        AbstractPower powerToApply = ReflectionHacks.getPrivate(__instance, ApplyPowerAction.class, "powerToApply");
+        if (p instanceof FrostprimePower && powerToApply instanceof FrostprimePower) {
+            ((FrostprimePower)p).stackAmount2(((FrostprimePower)p).amount2);
```
modargo, line 16: There's a bug with this -- it's double the stacks of the power right now since it pulls the amount to add from `p` instead of from `powerToApply`. 
```suggestion
            ((FrostprimePower)p).stackAmount2(((FrostprimePower)powerToApply).amount2);
```

src/main/java/thePackmaster/cards/contentcreatorpack/OlexasCloak.java
```diff
@@ -0,0 +1,30 @@
...
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class OlexasCloak extends AbstractContentCard {
+    public final static String ID = makeID("OlexasCloak");
+
+    public OlexasCloak() {
+        super(ID, -1, CardType.SKILL, CardRarity.COMMON, CardTarget.SELF);
+        baseBlock = 18;
+        PersistFields.setBaseValue(this, 2);
+        HoardField.setBaseValue(this, 3);
```
modargo, line 19: As part of looking at this PR I double-checked the Hoard keyword. Could you update that keyword (in Cardstrings.json for DimensionGatePack) to have #b before the two costs it mentions (X and 0), so they get highlighted properly?

src/main/resources/anniv5Resources/localization/eng/contentcreatorpack/Cardstrings.json
```diff
@@ -0,0 +1,64 @@
...
+    "DESCRIPTION": "${ModID}:MEGAthereal. NL Deal !D! damage."
+  },
+  "${ModID}:Wanderbots": {
+    "NAME": "Wanderbots",
+    "DESCRIPTION": "Gain !B! Block. NL Channel a ${ModID}:Wanderbot.",
+    "UPGRADE_DESCRIPTION": "${ModID}:Inscribed. NL Gain !B! Block. NL Channel a ${ModID}:Wanderbot."
+  },
+  "${ModID}:Retromation": {
+    "NAME": "Retromation",
+    "DESCRIPTION": "At the start of your turn, gain a card from a Pack not available to your draft. NL It gains Ethereal."
```
modargo, line 31: Minor and probably debatable, but I would phrase this as *in* your draft.
```suggestion
    "DESCRIPTION": "At the start of your turn, gain a card from a Pack not available in your draft. NL It gains Ethereal."
```

src/main/java/thePackmaster/powers/contentcreatorpack/RetromationPower.java
```diff
@@ -0,0 +1,74 @@
...
+    public void atStartOfTurnPostDraw() {
+        if (!potentialPacks.isEmpty()) {
+            this.flash();
+            for (int i = 0; i < amount; i++) {
+                ArrayList<String> potentialCardIDs = Wiz.getRandomItem(potentialPacks, AbstractDungeon.cardRandomRng).getCards();
+                ArrayList<AbstractCard> potentialCards = new ArrayList<>();
+                for (String s : potentialCardIDs) {
+                    AbstractCard test = CardLibrary.getCard(s).makeCopy();
+                    if (!test.hasTag(AbstractCard.CardTags.HEALING) && test.rarity != AbstractCard.CardRarity.SPECIAL)
+                        potentialCards.add(CardLibrary.getCard(s));
```
modargo, line 55: Not making a copy of the card here could cause issues, and switching this to add test to the test is a simple way to fix that (and shortens the code).
```suggestion
                        potentialCards.add(test);
```

src/main/java/thePackmaster/cards/contentcreatorpack/Rhapsody.java
```diff
@@ -0,0 +1,28 @@
...
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+import static thePackmaster.util.Wiz.applyToSelf;
+import static thePackmaster.util.Wiz.atb;
+
+public class Rhapsody extends AbstractContentCard {
+    public final static String ID = makeID("Rhapsody");
+    public Rhapsody() {
+        super(ID, 2, CardType.SKILL, CardRarity.RARE, CardTarget.SELF);
+        baseMagicNumber = magicNumber = 150;
```
modargo, line 17: That's a lot of inspiration -- 2.5x the effect of a damage or block card, or 3x the effect when upgraded. You don't get increases to secondary effects of the cards, so there are things that Burst/Double Tap/etc. can do that this can't, but this seems really high efficiency when you consider that it also draws 2 cards.

It also seems very high efficiency when you consider that bards 3-cost form power (Splendid Form) only gives 100 Inspiration at the start of each turn.

My gut reaction is that this should be toned down slightly -- don't need to overnerf it, just reduce the inspiration and/or draw slightly. It's fine to have powerful cards, so maybe the thing to do is to reduce the draw to 1. I think it'll still be a very good card (and incredible with Snecko Eye).

src/main/java/thePackmaster/patches/contentcreatorpack/PreDrawPatch.java
```diff
@@ -0,0 +1,19 @@
...
+
+public class PreDrawPatch {
+    public static boolean DRAWN_DURING_TURN = false;
+
+    @SpirePatch(
+            clz = AbstractPlayer.class,
+            method = "applyStartOfTurnRelics"
+    )
+    public static class AbstractPlayerApplyStartOfTurnRelicsPatch {
+        public static void Prefix(AbstractPlayer __instance) {
```
modargo, line 14: With how this is implemented, effects that draw additional cards at the start of your turn will count as draw for this card That's something I find unintuitive -- draw like that feels like part of your pre-turn draw to me, and that's where people have ended up with similar effects to this already in the Packmaster repo, such as the card Trade Beacon (see TradeBeaconPower for how it does things.

You can avoid this by targeting `atStartOfTurnPostDraw` instead, and having this addToBot an action that sets these flags (so that all the actions in the queue resolve, which will include additional draw at the start of the turn, and then these flags get set).

src/main/java/thePackmaster/cards/contentcreatorpack/Wanderbots.java
```diff
@@ -0,0 +1,42 @@
...
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        blck();
+        atb(new ChannelAction(new Wanderbot()));
+        if (upgraded) atb(new AbstractGameAction() {
+            @Override
+            public void update() {
+                isDone = true;
+                for (AbstractOrb o : AbstractDungeon.player.orbs) {
+                    if (!(o instanceof EmptyOrbSlot)) {
+                        o.onEndOfTurn();
```
modargo, line 33: Would it make sense to use `InscribedMod` instead of this code? This might have been written before that was in the repo, not sure.

Even if you don't use `InscribedMod`, you should copy over how it activates all different orbs types. Calling `onStartOfTurn` is needed for orbs like Plasma, and checking for `PackmasterOrb` and calling `passiveEffect` is necessary to support orbs that only have a passive effect (that is only triggered by cards like this, not start or end of turn).
```
        for (AbstractOrb o : AbstractDungeon.player.orbs){
            if (o instanceof PackmasterOrb){
                ((PackmasterOrb) o).passiveEffect();
            } else {
                o.onStartOfTurn();
                o.onEndOfTurn();
            }
        }
```
(The code above is pasted from `InscribedMod` and might benefit from a check for `EmptyOrbSlot`.)

src/main/java/thePackmaster/cards/highenergypack/Food.java
```diff
@@ -21,6 +22,7 @@ public Food() {
     }
 
     public void use(AbstractPlayer p, AbstractMonster m) {
+        atb(new TalkAction(true, cardStrings.EXTENDED_DESCRIPTION[4], 1F, 2F));
         atb(new GainEnergyAction(magicNumber));
```
modargo, line 26: Food is broken because it uses `magicNumber` here but never sets it. Right now it will give you -1 energy because of that.

src/main/java/thePackmaster/hats/HatMenu.java
```diff
@@ -82,7 +88,12 @@ public static void refreshHatDropdown() {
         dropdown = new DropdownMenu(((dropdownMenu, index, s) -> setCurrentHat(index, s)),
                 optionNames, FontHelper.tipBodyFont, Settings.CREAM_COLOR);
 
-        if (init) setCurrentHat(0, optionNames.get(0));
+
+        if (init) {
+            int lastPickedIdx = SpireAnniversary5Mod.getLastPickedHatIndex();
```
modargo, line 93: This is going to be unstable -- whenever we add another pack, indexes in the list will change.

What I suggest we do here is:
* Track the mapping of pack ID to dropdown index (in this class, when building the dropdown)
* Save the pack ID, not the index
* Have this code using the mapping to look up the right index

src/main/java/thePackmaster/orbs/contentcreatorpack/Wanderbot.java
```diff
@@ -0,0 +1,109 @@
...
+import thePackmaster.util.TexLoader;
+
+import static thePackmaster.util.Wiz.atb;
+
+public class Wanderbot extends AbstractOrb {
+    private static final OrbStrings orbString;
+    private static final Texture myTex = TexLoader.getTexture("anniv5Resources/images/orbs/wanderbot_orb.png");
+
+    public Wanderbot() {
+        this.ID = "Wanderbot";
```
modargo, line 36: This should use `makeID`.

While you're fixing this, I noticed that Ghostflame (which perhaps you copied this from?) also doesn't use makeID. Could you fix that as well?

src/main/java/thePackmaster/packs/DefectPack.java
```diff
@@ -26,9 +26,9 @@ public ArrayList<String> getCards() {
         cards.add(Hologram.ID);
         cards.add(Defragment.ID);
         cards.add(Sunder.ID);
-        cards.add(Chaos.ID);
+        cards.add(GoForTheEyes.ID);
         cards.add(Fusion.ID);
-        cards.add(Skim.ID);
```
modargo, line 31: Noooooooooooooooooooooooooooooooooooooooooooooooooo my card draw.

It's fine, just that I like Skim. Most other packs have enough card draw in them anyway.

src/main/java/thePackmaster/patches/contentcreatorpack/OnDrawCard.java
```diff
@@ -0,0 +1,34 @@
+package thePackmaster.patches.contentcreatorpack;
+
+import com.evacipated.cardcrawl.modthespire.lib.*;
+import com.megacrit.cardcrawl.cards.AbstractCard;
+import com.megacrit.cardcrawl.characters.AbstractPlayer;
+import javassist.CtBehavior;
+
+public class OnDrawCard {
```
modargo, line 8: Might be more accurate to call this `TrackCardsDrawnDuringTurnPatch`. I'm commenting on this most so that some future contributor searching for OnDraw is less likely to get confused -- with the current name this class sounds like a type of AbstractCard (one of the interfaces you would implement) and sounds like it would trigger for any draw at all.

src/main/java/thePackmaster/patches/contentcreatorpack/VoyagedPatch.java
```diff
@@ -0,0 +1,19 @@
...
+
+public class VoyagedPatch {
+
+    @SpirePatch(
+            clz = AbstractPlayer.class,
+            method = "applyStartOfTurnPostDrawRelics"
+    )
+    public static class AbstractPlayerApplyStartOfTurnPostDrawRelicsPatch {
+        public static void Prefix(AbstractPlayer __instance) {
+            AbstractDungeon.actionManager.addToBottom(new EnableDrawnDuringTurnAction());
```
modargo, line 16: It looks like this patch and PreDrawPatch are doing the same thing, but in slightly different ways? This is the way (adding an action after one of the post draw methods) that I think works better and described in my comment on PreDrawPatch, so maybe just delete the current PreDrawPatch entirely and rely on this patch? (Maybe renamed to not use the adventurer-specific Voyaged term.)

Let me know if I'm missing something.

src/main/resources/anniv5Resources/localization/eng/contentcreatorpack/Powerstrings.json
```diff
@@ -0,0 +1,44 @@
...
+      " cards next turn."
+    ]
+  },
+  "${ModID}:RetromationPower": {
+    "NAME": "Retromation",
+    "DESCRIPTIONS": [
+      "At the start of your turn, gain #b",
+      " card",
+      " cards",
+      " from a Pack not available to your draft. NL ",
```
modargo, line 23: Same common as for the card, maybe use "in".
```suggestion
      " from a Pack not available in your draft. NL ",
```

src/main/resources/anniv5Resources/localization/eng/highenergypack/Cardstrings.json
```diff
@@ -32,7 +32,8 @@
       "Gain ",
       "Retain. NL Gain ",
       " . NL Exhaust.",
-      "You gain nothing! NL Good day, sir! NL Exhaust."
+      "You gain nothing! NL Good day, sir! NL Exhaust.",
+      "mm delocious"
```
modargo, line 36: I'm going to assume the spelling mistakes are intentional.

src/main/resources/anniv5Resources/localization/eng/odditiespack/UIstrings.json
```diff
@@ -8,8 +8,8 @@
   },
   "${ModID}:OdditiesPackHat": {
     "TEXT": [
-      "Temp!.",
-      "Temp Hat"
+      "good for Adventuring.",
```
modargo, line 11:
```suggestion
      "Good for Adventuring.",
```

modargo (changes requested):
Cool pack! I tried to go through in detail with comments so that this pack can be in great shape, since I'm sure it will get some attention (and loaded it up in game to see it in action). Hope the comments all make sense and are helpful.

I also added comments for the other things that got lumped in with this review -- it might be best to separate these things in the future, to help organize and prioritize better. You can do that either by moving one-off fixes to multiple branches, or by creating more incremental PRs (so that you can push small fixes first and get those reviewed/merged before bigger stuff). As it stands, though, we'll just need to get everything in the PR in shape to merge.

src/main/resources/anniv5Resources/localization/eng/contentcreatorpack/Cardstrings.json
```diff
@@ -0,0 +1,64 @@
...
+    "DESCRIPTION": "At the start of your turn, Channel 1 Frost and apply !M! ${ModID}:Frostbite to ALL enemies."
+  },
+  "${ModID}:BaalorBash": {
+    "NAME": "Baalorlord's Bash",
+    "DESCRIPTION": "Deal !D! damage. NL Shuffle a *Baalor's *Blueprint into your draw pile. NL Exhaust.",
+    "UPGRADE_DESCRIPTION": "Innate. NL Deal !D! damage. NL Shuffle a *Baalor's *Blueprint into your draw pile. NL Exhaust."
+  },
+  "${ModID}:BaalorBlueprint": {
+    "NAME": "Baalor's Blueprint",
+    "DESCRIPTION": "Shuffle your discard pile into your draw pile. Put *The *Cozy *Chair on the bottom of it. Exhaust."
```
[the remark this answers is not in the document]
mikemayhemdev: This is intentionally not on a new line to prevent the 6-line break.  There's rare precedence in base game for this.

src/main/java/thePackmaster/cards/contentcreatorpack/AbstractContentCard.java
```diff
@@ -0,0 +1,17 @@
+package thePackmaster.cards.contentcreatorpack;
+
+import thePackmaster.cards.AbstractPackmasterCard;
+
+public abstract class AbstractContentCard extends AbstractPackmasterCard {
+
+    public boolean reroll = false, isChance = false;
```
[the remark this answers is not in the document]
mikemayhemdev: Yeah looks like copypasta from boardgames.

src/main/java/thePackmaster/cards/contentcreatorpack/TheCozyChair.java
```diff
@@ -0,0 +1,25 @@
...
+
+import com.megacrit.cardcrawl.characters.AbstractPlayer;
+import com.megacrit.cardcrawl.monsters.AbstractMonster;
+import thePackmaster.cards.AbstractPackmasterCard;
+import thePackmaster.powers.contentcreatorpack.CozyChairPower;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+import static thePackmaster.util.Wiz.applyToSelf;
+
+public class TheCozyChair extends AbstractContentCard {
```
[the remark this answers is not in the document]
mikemayhemdev: Worth noting that the original card doesn't trigger an upgrade chain, those upgrades are only if you do something like an Armaments effect.

I feel like there's big risk if we make the payoff card Ethereal there's big risk to you doing the quest and it getting burned.  4-cost might work, but then I think we have to do some kind of double-card-preview shenanigans to tell you its unplayable without assistance.

src/main/java/thePackmaster/cards/contentcreatorpack/BaalorBash.java
```diff
@@ -0,0 +1,29 @@
...
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class BaalorBash extends AbstractContentCard {
+    public final static String ID = makeID("BaalorBash");
+
+    public BaalorBash() {
+        super(ID, 1, CardType.ATTACK, CardRarity.RARE, CardTarget.ENEMY);
+        baseDamage = 13;
+        exhaust = true;
+        cardsToPreview = new BaalorBlueprint();
```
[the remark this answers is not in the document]
mikemayhemdev: This was an intentional choice to keep the effect a mystery, the same as Omega does. I personally enjoy having these weird side quest (obviously meme) cards have that, but I understand if we expect players at large do not.

src/main/java/thePackmaster/patches/contentcreatorpack/OnDrawCard.java
mikemayhemdev: Agreed

src/main/java/thePackmaster/patches/contentcreatorpack/VoyagedPatch.java
mikemayhemdev: Yeah this also looks suspicious, i'll try with your suggestion and see what happens.

mikemayhemdev: Actually i'll let @DarkVexon take care of this because I feel this is probably some weird edge case that I won't find in a test.

src/main/resources/anniv5Resources/localization/eng/contentcreatorpack/Cardstrings.json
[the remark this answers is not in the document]
modargo: I've usually felt that lack of a line break looks odder than the extra line, but I agree that it's a something where you could reasonably go the other way.

src/main/java/thePackmaster/cards/contentcreatorpack/BaalorBash.java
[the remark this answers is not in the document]
modargo: Personally I'm fine with releasing it either way -- we can keep an air of mystery if we want :smile:. Multi-card preview is a pretty new thing that I doubt players even expect to have.

src/main/java/thePackmaster/cards/contentcreatorpack/AbstractContentCard.java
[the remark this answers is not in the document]
DarkVexon: Removed!

src/main/java/thePackmaster/cards/contentcreatorpack/OlexasCloak.java
DarkVexon: Done!

src/main/java/thePackmaster/patches/contentcreatorpack/PreDrawPatch.java
DarkVexon: I believe that's what gets done currently? The VoyagedPatch adds an action at post draw which then enables the effect. This Pre-Draw patch is what sets the flag to false before all the draw is incurred.

src/main/java/thePackmaster/cards/contentcreatorpack/Wanderbots.java
DarkVexon: I'll swap it to use InscribedMod for consistency, thanks!

DarkVexon: Also added a check for emptyorbslot in the InscribedMod and added Inherent functionality.

src/main/java/thePackmaster/orbs/contentcreatorpack/Wanderbot.java
DarkVexon: Copy that, done!

src/main/java/thePackmaster/patches/contentcreatorpack/OnDrawCard.java
DarkVexon: Done!

src/main/java/thePackmaster/patches/contentcreatorpack/VoyagedPatch.java
DarkVexon: The three patches work together. One disables counting cards at the start of the turn, the other adds the action that re-enables it, and the OnDrawCard/TrackCardsDrawnDuringTurnPatch is what triggers on draw, checks against the variable and actually counts the drawn card. I've renamed the classes so hopefully it's a little more clear.

src/main/java/thePackmaster/hats/HatMenu.java
DarkVexon: hmm - I started on this, but that actually doesn't work with the Randomized Hat option right now since that one isn't attached to an ID in the hats and only exists in the options

DarkVexon: OK, got it

src/main/java/thePackmaster/cards/highenergypack/Food.java
modargo: Interesting -- be aware that causes it to give you -1 energy if you try to test it by, for example, add it from the console.

src/main/java/thePackmaster/patches/contentcreatorpack/VoyagedPatch.java
modargo: Thanks. I understand what you're saying -- having the code spread out across multiple patches made it hard to put everything together. It would be a lot more readable if all the patches were consolidated into one place (in addition to the naming changes you mentioned).

PR #229, opened 2023-01-18
Startup -> GrandOpening and BasicsPack

GameboyJ:
some changes to startup, including a new name

new pack based around basic cards


erasels:
Hey, I really appreciate you contributing to the project, however I have to be the bearer of bad news.
With the quantity of packs we have in the project at this point it's important that we take a closer look at the packs and sadly the basics pack falls a little short of the required design standards.

The pack boils down to a combination of the base classes with none of their interesting mechanics mixed with a lot of screens. I've seen an archetype like this in Hermit which had genuinely creative effects, so it can definitely be done. 

Long story short, I won't be merging this pack in its current state, sorry. If you really want to play with the idea some more, I'd recommend taking a look at Hermit and a critical look at what the pack wants to achieve while also keeping in mind that you need some form of scaling.

PR #232, opened 2023-01-18
Final Frames!

mikemayhemdev:


mikemayhemdev:
Slid in a fix for Cheapen hitting X-costs into this.  It simply can't now, but if I remember there's some weird thing where "card is free" + "x cost" = "play for free at current X"?  Would someone who knows about that expect Cheapen hitting an X-cost to behave that way?  If so, might need some assistance knowing the right fields to modify.

PR #235, opened 2023-01-18
-open bracket rework

AlexMdle:


modargo:
Looks good, and I like the new design for this card! It's a neat way to make 0-cost draw that isn't repeatable on the same turn.

PR #236, opened 2023-01-18
RimWorld Art!

Laugic:
Added credits
Removed decompiler line comment
Make cards inherit abstract card for card back


modargo:
The art looks very cool! Thanks for getting this in place ahead of release.

PR #254, opened 2023-01-18
startup -> grandopening

GameboyJ:
renaming all startup stuff to grandopening, to prevent confusion with the startup mechanic and lean more into the theatre flavour of the pack


erasels:
Please don#t remove exmaple pom

erasels (changes requested):

PR #270, opened 2023-01-19
Packmaster Character Event (and a few fixes)

mikemayhemdev:


src/main/java/thePackmaster/events/BlackMarketDealerEvent.java
```diff
@@ -0,0 +1,174 @@
...
+import com.megacrit.cardcrawl.localization.EventStrings;
+import com.megacrit.cardcrawl.vfx.cardManip.ShowCardAndObtainEffect;
+import thePackmaster.SpireAnniversary5Mod;
+
+import java.util.Iterator;
+
+import static BattleTowers.BattleTowers.makeID;
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class BlackMarketDealerEvent extends AbstractImageEvent {
```
modargo, line 26: Was this a WIP version of an event to get Pack Rip or something like that? Looks to me like it shouldn't be part of this PR -- everything else looks good, so if you remove this, I can merge the rest.

mikemayhemdev: Ah crap, I thought I was on another branch already.  Well I just pushed the full event, so this is now "two fixes and a giant event" PR.

PR #283, opened 2023-01-19
Even/Odd pack

lobbienjonsji:
It has finally been done! All that's missing is the hat.


src/main/java/thePackmaster/actions/evenoddpack/BattleCryAction.java
```diff
@@ -0,0 +1,23 @@
...
+import com.megacrit.cardcrawl.powers.StrengthPower;
+import thePackmaster.util.Wiz;
+
+public class BattleCryAction extends AbstractGameAction {
+    
+    public BattleCryAction(int amount) {
+        this.amount = amount;
+    }
+    
+    public void update() {
```
erasels, line 16: Please make this an anonymous action, they help centralizing the logic and keep the project easier to oversee for me

src/main/java/thePackmaster/actions/evenoddpack/ElementOfSurpriseAction.java
```diff
@@ -0,0 +1,20 @@
+package thePackmaster.actions.evenoddpack;
+
+import com.megacrit.cardcrawl.actions.AbstractGameAction;
+import com.megacrit.cardcrawl.actions.common.GainBlockAction;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.powers.DuplicationPower;
+import thePackmaster.util.Wiz;
+
+public class ElementOfSurpriseAction extends AbstractGameAction {
```
erasels, line 9: anonymous action candidate

src/main/java/thePackmaster/actions/evenoddpack/FlourishAction.java
```diff
@@ -0,0 +1,28 @@
...
+import com.megacrit.cardcrawl.actions.AbstractGameAction;
+import com.megacrit.cardcrawl.actions.common.DamageAction;
+import com.megacrit.cardcrawl.actions.common.GainBlockAction;
+import com.megacrit.cardcrawl.actions.common.GainEnergyAction;
+import com.megacrit.cardcrawl.actions.common.HealAction;
+import com.megacrit.cardcrawl.cards.DamageInfo;
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+
+public class FlourishAction extends AbstractGameAction {
```
erasels, line 12: Anotha one

src/main/java/thePackmaster/actions/evenoddpack/GammaWardAction.java
```diff
@@ -0,0 +1,28 @@
...
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.monsters.AbstractMonster;
+import com.megacrit.cardcrawl.powers.VulnerablePower;
+import com.megacrit.cardcrawl.powers.WeakPower;
+import com.megacrit.cardcrawl.vfx.StarBounceEffect;
+import thePackmaster.powers.evenoddpack.GammaWardPower;
+import thePackmaster.powers.evenoddpack.PrimeDirectivePower;
+import thePackmaster.util.Wiz;
+
+public class GammaWardAction extends AbstractGameAction {
```
erasels, line 16: We ask, again

src/main/java/thePackmaster/actions/evenoddpack/QuickReflexAction.java
```diff
@@ -0,0 +1,24 @@
+package thePackmaster.actions.evenoddpack;
+
+import com.megacrit.cardcrawl.actions.AbstractGameAction;
+import com.megacrit.cardcrawl.actions.common.DamageAction;
+import com.megacrit.cardcrawl.actions.common.GainBlockAction;
+import com.megacrit.cardcrawl.cards.DamageInfo;
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+
+public class QuickReflexAction extends AbstractGameAction {
```
erasels, line 10: Anonymous please

src/main/java/thePackmaster/actions/evenoddpack/SwordAndBoardAction.java
```diff
@@ -0,0 +1,32 @@
...
+
+import com.megacrit.cardcrawl.actions.AbstractGameAction;
+import com.megacrit.cardcrawl.actions.common.DamageAction;
+import com.megacrit.cardcrawl.actions.common.DrawCardAction;
+import com.megacrit.cardcrawl.actions.common.GainBlockAction;
+import com.megacrit.cardcrawl.cards.DamageInfo;
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+
+public class SwordAndBoardAction extends AbstractGameAction {
```
erasels, line 11: You know what I'll ask

src/main/java/thePackmaster/actions/evenoddpack/WindupAction.java
```diff
@@ -0,0 +1,34 @@
...
+import com.megacrit.cardcrawl.actions.AbstractGameAction;
+import com.megacrit.cardcrawl.actions.common.DamageAction;
+import com.megacrit.cardcrawl.actions.common.GainBlockAction;
+import com.megacrit.cardcrawl.actions.common.ModifyDamageAction;
+import com.megacrit.cardcrawl.cards.AbstractCard;
+import com.megacrit.cardcrawl.cards.DamageInfo;
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+
+public class WindupAction extends AbstractGameAction {
```
erasels, line 12: I'll stop pointing them out. Basically if it's a one-off action and has less than 10 lines, please make it an anonymous action

src/main/java/thePackmaster/patches/evenoddpack/AddAuraTextPatch.java
```diff
@@ -0,0 +1,19 @@
...
+
+@SpirePatch(
+        clz = AbstractCard.class,
+        method = SpirePatch.CONSTRUCTOR,
+        paramtypez = {String.class, String.class, String.class, int.class, String.class, AbstractCard.CardType.class, AbstractCard.CardColor.class, AbstractCard.CardRarity.class, AbstractCard.CardTarget.class, DamageInfo.DamageType.class}
+)
+public class AddAuraTextPatch {
+    @SpirePostfixPatch
+    public static void AddAuraText(AbstractCard __instance, String id, String name, String imgUrl, int cost, String rawDescription, AbstractCard.CardType type, AbstractCard.CardColor color, AbstractCard.CardRarity rarity, AbstractCard.CardTarget target, DamageInfo.DamageType dType) {
+        CardModifierManager.addModifier(__instance, new AuraTextModifier());
```
erasels, line 17: This is kind of bad. I'd really prefer it if this only existed when necessary. Can't you apply this to all cards when the power activates or something? It's a lot of additional logic for every card

erasels (changes requested):

src/main/java/thePackmaster/patches/evenoddpack/AddAuraTextPatch.java
lobbienjonsji: I can give it a try, but I can't promise to find a better solution without it having some sort of edge case where the Cardmodifier won't get removed from/added to the card. What I could do and really should have done, now that I think of it, is, instead of meddling with the cards applyPowers method I could simply make the power only call initializeDescription for all cards in the players hand, since any other card should not have their card text updated anyways. That way there are a lot fewer calls to initializeDescription and I could significantly simplify the logic modify description as well, by simply putting all the logic in the power instead.

lobbienjonsji: First off: If there was a sort of afterApplyPowers hook, that passes the card on which the method was called, I would not need cardmods at all, since i could simply add the necessary card text afterwards (It NEEDS to happen AFTER apply powers, since some cards modify their description in applyPowers).
I could attempt making one, but I'd prefer not to.

To answer the question: I'd have to find a way to apply the effect to cards that might get put into the hand between effect triggers.
The only place I know where to check for new cards in hand consistently, that isn't an update method (which seems like a terrible place to check for new cards in hand), would be the cards own applyPowers method.
(If you know any other ways, let me know). 
I would patch applyPowers to check for the power existing, if it is there, apply the cardmod to itself.
This would put half the logic in the patch instead, but at least the modifyDescription part of the cardmod would only exist for cards that are in the players hand while they have the power.
I'd also extend the ResetAttributesPatch to also remove the cardmod again.

erasels: >(If you know any other ways, let me know).
`CardGroup:refreshHandLayout()` is pretty consistent for catching changes to the hand. It gets called by most methods that add or remove a card to the hand and a bunch in general. It's what Spinning Top uses to detect no more cards being in the hand.

I think a combination of postfixing that to apply it to everything in hand and removing it (or not, I don't particularly care about that here) would work pretty well. Maybe also call the code once in `onInitialApplication()` of the power. That should cover the cases I can think of.

In case that doesn't work for some reason I'd also be fine with literally adding a static field for the power that's true when the power is in play and false when it's not. Prefix the logic checks of the modifier with that and it'd be alright if not especially elegant. (my main gripe is that every card will always check for hasPower, it's just a bit much unneeded iteration for my tastes)

erasels (approved):

PR #290, opened 2023-01-20
Core set relics: fix PM Booster Box to properly save/load and fix both it and Banishing Decree to use seeded RNG and respect relics like Question Card and Busted Crown for their rewards

modargo:
The core idea of this implementation is that the way to properly save and load a card reward that has custom logic is to use a custom reward. Shouldn't be too surprising, but the tricky part is properly implementing custom rewards that generate cards and don't do weird things with RNG requires extra work.

That's because rewards are added right before the end-of-combat autosave, so any RNG calls advance the RNG to a new state that is then saved. If you save/reload, the reward is recreated, and it will advance RNG again. This solution avoids that by only generating the cards in the reward after the end-of-combat autosave -- this actually follows what the base game does with card rewards (the reward is only generated when you open the rewards screen).

There are definitely other ways to do this, but as far as I'm aware you have to do something along these lines to avoid RNG instability on save/load. This particular solution is based on what I did for Corrupt the Spire, so the core idea has been heavily tested and it's just the specific adaptation here that might introduce issues. See the following files for a sampling of the Corrupt the Spire implementation:
* https://github.com/modargo/corruptthespire/blob/f47f2a7b9ee5950e128d01960d519c6ed949fb7d/src/main/java/corruptthespire/rewards/CorruptedCardReward.java
* https://github.com/modargo/corruptthespire/blob/f47f2a7b9ee5950e128d01960d519c6ed949fb7d/src/main/java/corruptthespire/patches/fight/FightCorruptionAddRewardsPatch.java
* https://github.com/modargo/corruptthespire/blob/f47f2a7b9ee5950e128d01960d519c6ed949fb7d/src/main/java/corruptthespire/patches/rewards/PopulateCardsForCorruptedCardRewardPatch.java

### DIFF
--- a/src/main/java/thePackmaster/SpireAnniversary5Mod.java
+++ b/src/main/java/thePackmaster/SpireAnniversary5Mod.java
@@ -31,20 +31,18 @@ import com.megacrit.cardcrawl.core.AbstractCreature;
 import com.megacrit.cardcrawl.core.CardCrawlGame;
 import com.megacrit.cardcrawl.core.Settings;
 import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
-import com.megacrit.cardcrawl.events.city.BackToBasics;
 import com.megacrit.cardcrawl.helpers.CardHelper;
 import com.megacrit.cardcrawl.helpers.CardLibrary;
 import com.megacrit.cardcrawl.helpers.FontHelper;
 import com.megacrit.cardcrawl.localization.*;
-import com.megacrit.cardcrawl.orbs.*;
 import com.megacrit.cardcrawl.powers.AbstractPower;
 import com.megacrit.cardcrawl.powers.ArtifactPower;
 import com.megacrit.cardcrawl.powers.watcher.VigorPower;
 import com.megacrit.cardcrawl.random.Random;
+import com.megacrit.cardcrawl.rewards.RewardSave;
 import com.megacrit.cardcrawl.rooms.AbstractRoom;
 import com.megacrit.cardcrawl.stances.AbstractStance;
 import com.megacrit.cardcrawl.stances.CalmStance;
-import com.megacrit.cardcrawl.stances.NeutralStance;
 import com.megacrit.cardcrawl.unlock.UnlockTracker;
 import javassist.CtClass;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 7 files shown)


src/main/java/thePackmaster/SpireAnniversary5Mod.java
```diff
@@ -707,24 +703,22 @@ public static AbstractCard getRandomCardFromPack(AbstractCardPack pack) {
...
-    public static ArrayList<AbstractCard> getCardsFromPacks(String pack, int count) {
+    public static ArrayList<AbstractCard> getCardsFromPacks(String pack, int count, Random rng) {
         ArrayList<String> quick = new ArrayList<>();
         quick.add(pack);
-        return getCardsFromPacks(quick, count);
+        return getCardsFromPacks(quick, count, rng);
     }
 
-    public static ArrayList<AbstractCard> getCardsFromPacks(ArrayList<String> packs, int count) {
+    public static ArrayList<AbstractCard> getCardsFromPacks(ArrayList<String> packs, int count, Random rng) {
```
modargo, line 712: Passing the rng here (and in the method above) is the key change. There were also some formatting changes because the code was formatting in a really odd way that I wanted to improve.

src/main/java/thePackmaster/relics/BanishingDecree.java
```diff
@@ -118,9 +119,14 @@ public void update() {
                 }
 
                 for (int i = 0; i < 3; i++) {
-                    RewardItem r = new RewardItem();
-                    r.cards = getCardsFromPacks(cp.packID, 3);
-                    AbstractDungeon.getCurrRoom().rewards.add(r);
+                    RewardItem reward = new RewardItem();
+                    reward.cards = getCardsFromPacks(cp.packID, reward.cards.size(), AbstractDungeon.cardRng);
```
modargo, line 123: What the `reward.cards.size()` on this line (and the similar one in PMBoosterBoxCardReward) is doing:

The base game has two ways that the number of cards in a reward can be modified (relics like Question Card/Busted Crown and the custom run modifier Binary). You could simply reproduce that logic to determine the number of cards to have in a reward, and that would work fine.

However, what if a mod adds some new thing that modifiers card rewards? They'll have patched the logic in `AbstractDungeon.getRewardCards`, but if you've made your own method that has the same logic, there's no way some other mod will know about and patch that method.

So since we're a bit paranoid, we generate a RewardItem and check how many cards there are, and then just use that number for our custom card reward.

On a separate note, the card reward for Banishing Decree does not need to be a registered custom card reward, because there's no need for it to save/load. If you reload after getting Banishing Decree but before going to the next room, you'll need to pick up Banishing Decree again and that will create the card reward again.

PR #331, opened 2023-01-23
show pack name in singlecardview

OceanUwU:
image

### DIFF
--- a/src/main/java/thePackmaster/patches/SingleCardViewBorderTextPatch.java
+++ b/src/main/java/thePackmaster/patches/SingleCardViewBorderTextPatch.java
@@ -1,11 +1,18 @@
 package thePackmaster.patches;
 
+import com.badlogic.gdx.graphics.Color;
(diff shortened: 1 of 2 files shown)


src/main/java/thePackmaster/patches/SingleCardViewBorderTextPatch.java
```diff
@@ -0,0 +1,40 @@
...
+            yField.setAccessible(true);
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+    }
+
+    public static void Postfix(SingleCardViewPopup __instance, SpriteBatch sb) {
+        try {
+            AbstractCard card = (AbstractCard)cardField.get(__instance);
+            if (card instanceof AbstractPackmasterCard) {
```
erasels, line 33: This will only work on custom cards, base game cards would not show the text. 
Look at `RenderBaseGameCardPackTopTextPatches`

erasels (changes requested):

PR #341, opened 2023-01-24
Jeweler's Pack (Unique submission, needs rule review)

mikemayhemdev:
So this is a very different kind of pack - a Pure Support pack of all unplayable Gems which operate almost identically to Guardian's gem/socket mechanic (now using cardmods instead of Guardians' very jank system).

In testing the pack, Common gems proved a little difficult in the context of PM as they can too easily and reliably escalate other cards.  So, I'm proposing that this pack is 5 Uncommons and 5 Rares given its unique position and structure, an exception to the usual rule.

There is also a consideration that every card in the pack has the HEALING tag, which means this pack is exempted from all in-combat generation that is based on packs.  Booster Brew, Destiny Draw, etc. all have exception cases where they cannot pick the GemsPack due to its unique behaviors.

If these two exceptions are more than we're comfortable with, I'm happy to go back to the drawing board and shelve this pack for now, or implement some alternatives to get it back into the ruleset:

For the common issue, moving some gems to Common and increasing the gold cost to socket (to perhaps 25g) would probably be sufficient to make the commons not pickable simply due to gold availability.  Even if they appear strong, the limitations surrounding them would make them less draftable, though this wouldn't be easily apparent on first use of the pack.

The instant alternative to the second exception would be to make the Gems playable in combat, but I intended to have them be a draft pick that is a dead draw until it is socketed, so there is more inherent risk to picking a lot of them up - it adds a map strategy element to using the pack.


mikemayhemdev:
Misclicked when creating this PR, intended to click the arrow next to it to see if it could be saved for later.  But its here now so feel free to review and/or comment.

PR #345, opened 2023-01-24
Some small buffs & nerfs

mikemayhemdev:
Block VIal+ now is 6 block next turn instead of blur.

Talk to the Claw+ now 1 less block generation.

Lethal Shot - +2 Poison

Knighthood - Costs 3, but upgrades to gain retain.

Inferno - Base cost reduced to 8, upgrade takes it to 6.

Scorched Earth -Damage +1


mikemayhemdev:
This also now includes the base game pack changes discussed in discord.

PR #349, opened 2023-01-24
Warlock pack: fix Imps being stuck in your hand if a card that opens up a hand selection screen is played before the Imps trigger

modargo:


src/main/java/thePackmaster/cards/warlockpack/Imp.java
```diff
@@ -30,7 +30,6 @@ public Imp() {
         super(ID, COST, AbstractCard.CardType.SKILL, AbstractCard.CardRarity.SPECIAL, AbstractCard.CardTarget.ALL_ENEMY, CardColor.COLORLESS);
         magicNumber = baseMagicNumber = DAMAGE;
         secondDamage = baseSecondMagic = DAMAGE_INCREASE;
-        AutoplayField.autoplay.set(this, true);
```
modargo, line 33: This was just cleaning up something that was redundant and confusing my debugging -- the patches I wrote handle everything Autoplay does, and having the autoplay set was redundant (I noticed it because it was trying to add Imps to the action queue twice, but the second time it would detect that the card was already in the queue).

src/main/java/thePackmaster/patches/warlockpack/ImpGameActionManagerPatch.java
```diff
@@ -0,0 +1,33 @@
...
+
+// HandCardSelectScreen calls GameActionManager.cleanCardQueue, and is one of only two places that calls that method, the
+// other being AbstractMonster.die(). If you draw imps at the same time as a card that opens HandCardSelectScreen is played,
+// this means that the card queue item for playing the imps vanishes. To fix this, we add them back in an action that
+// executes afterwards.
+@SpirePatch(clz = GameActionManager.class, method = "cleanCardQueue", paramtypez = {})
+public class ImpGameActionManagerPatch {
+    @SpirePostfixPatch
+    public static void addImpsBackToQueue() {
+        if (AbstractDungeon.currMapNode != null && AbstractDungeon.currMapNode.room != null && AbstractDungeon.currMapNode.room.monsters != null && !AbstractDungeon.currMapNode.room.monsters.areMonstersBasicallyDead()) {
```
modargo, line 19: This series of checks may be more paranoid than necessary, but the core reason for having it is for the `areMonstersBasicallyDead` check at the end. That's because `cleanCardQueue` is called from `AbstractMonster.die` when `areMonstersBasicallyDead` is `true`, and in that case, we don't want to add imps back to the queue (since combat is over).

src/main/java/thePackmaster/patches/warlockpack/ImpGameActionManagerPatch.java
```diff
@@ -0,0 +1,33 @@
...
+            AbstractDungeon.actionManager.addToBottom(new AbstractGameAction() {
+                @Override
+                public void update() {
+                    for (AbstractCard c : AbstractDungeon.player.hand.group) {
+                        if (c instanceof Imp) {
+                            ((Imp)c).triggerWhenAddedToHand();
```
modargo, line 25: This is the same function that gets called normally for triggering Imps, and after it does various other stuff it will add an item to the card queue for the Imp.

PR #357, opened 2023-01-26
Initial ZHS Introduction

Mwallx:
Sometimes I change a bit of code for text, thought probably better off updating them routinely to the master than fixing them all at a final massive PR.


src/main/resources/anniv5Resources/localization/zhs/boardgamepack/Cardstrings.json
```diff
@@ -0,0 +1,53 @@
+{
+  "${ModID}:RollToMove": {
+    "NAME": "Roll to Move",
+    "DESCRIPTION": "Roll !M! ${ModID}:D6。"
+  },
+  "${ModID}:FlickDice": {
+  "NAME": "Flick Dice",
+  "DESCRIPTION": "{@@}造成 !D! 点伤害。 NL Roll a ${ModID}:D6。",
```
erasels, line 8: If you remove the dynamic text changes from the card, you don't have to use the tag anymore.

erasels (commented):
Please don't push the example-pom.xml removal to your PR. As a local change this is fine

src/main/resources/anniv5Resources/localization/zhs/boardgamepack/Cardstrings.json
Mwallx: This is just a result of several massive replacement. There is no need for dynamic text for ZHS, they will be removed once my progress gets here. For the pom thing, yesterday my change to it was tracked despite it being in .gitignore, so I searched some code and got rid of it. You could accept the PR and add it back manually, future PR shouldn't have this issue I suppose.

PR #359, opened 2023-01-26
Heavy is the head that wears the Clawn...

mikemayhemdev:
Claw Pack gutted... by making Ghost Claws not inherit Sharpen scaling.  Bumped up their damage by 1 and Sharpen by 1 so they match the original Claw card instead of being a weaker Claw, but yeah.  These are the main contributor to clawPack scaling too hard.  It should require some effort to scale up, and encourage proper use of Cardistry and other recursion effects and card draw rather than Claw Pack being completely self-sufficient.

Fixed the ClawUp method increasing base damage and flashing in red Claws that don't deal damage.

Alclawmize's generate claws potion now gives normal Claws instead of Ghost Claws.

Also simplified the Hoard keyword significantly, no more Persist or Retain, that effect is all baked into the Hoard keyword itself.  It now works with non-X cost cards, and both Olexa's Cloak and Inferno now cost 1.  (Kept the X-Cost functionality should someone else decide to use it).  Each play simply reduces the Hoard value by the card's cost (1 most of the time, but this way it isn't completely gutted by getting snecko'd).  The card can be repeated indefinitely to keep reducing the Hoard value until triggered.

Olexa Cloak - +2 Block.

### DIFF
--- a/src/main/resources/anniv5Resources/localization/eng/dimensiongatepack/Keywordstrings.json
+++ b/src/main/resources/anniv5Resources/localization/eng/dimensiongatepack/Keywordstrings.json
@@ -4,6 +4,6 @@
     "NAMES": [
       "hoard"
     ],
-    "DESCRIPTION": "When played, reduce Hoard by #bX (the card's cost). When Hoard reaches #b0, its effects trigger."
+    "DESCRIPTION": "When played, if this card's cost is greater than 0 and lower than it's Hoard: NL Reduce Hoard by the card's cost and ignore all other card effects. NL Return the card to your hand. NL The card #yRetains this turn."
   }
 ]
--- a/src/main/java/thePackmaster/SpireAnniversary5Mod.java
+++ b/src/main/java/thePackmaster/SpireAnniversary5Mod.java
@@ -231,7 +231,6 @@ public class SpireAnniversary5Mod implements
     public static boolean selectedCards = false;
     public static int combatExhausts = 0;
 
-    public static int CLAW_SHARP_TRACKER = 0;
 
     public static String makeID(String idText) {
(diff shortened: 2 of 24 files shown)


src/main/resources/anniv5Resources/localization/eng/dimensiongatepack/Keywordstrings.json
```diff
@@ -4,6 +4,6 @@
     "NAMES": [
       "hoard"
     ],
-    "DESCRIPTION": "When played, reduce Hoard by #bX (the card's cost). When Hoard reaches #b0, its effects trigger."
+    "DESCRIPTION": "When played, if this card's cost is greater than 0 and lower than it's Hoard: NL Reduce Hoard by the card's cost and ignore all other card effects. NL Return the card to your hand. NL The card #yRetains this turn."
```
erasels, line 7: I do not like this iteration of the keyword at all. It's a mechanics overload that tries to cover all angles but makes the card less parseable as a result.

I definitely understand why these changes were made after you watched Frost being confused but I think you have to keep in mind that he's dyslexic and a streamer (baiting donations, looking at chat instead of taking a second, etc)

Retain and Persist are disjointed here and should not be part of this already massive keyword. If you can't stomach putting those keyword on the card due to complexity, then maybe the effect need a bit more of an overhaul.
Also, the new wording o the base effect is even more confusing than it's first iteration `if this card's cost is greater than 0 and lower than it's Hoard` I definitely needed a few rereads to get this because by covering all angles you made the sentence hard to parse.
Simplicity is important and this really misses the mark.

erasels (changes requested):

erasels (approved):

PR #377, opened 2023-01-27
Nerfed Recall but also made it stackable

ActuallyEnbeon:
Hopefully should make the power's true nature easier to understand for players


ActuallyEnbeon:
Please don't merge this yet!! I found a bug!

src/main/resources/anniv5Resources/localization/eng/cosmoscommandpack/Powerstrings.json
```diff
@@ -10,7 +10,10 @@
   "${ModID}:RecallPower": {
     "NAME": "Recall",
     "DESCRIPTIONS": [
-      "Whenever a card is #yAmplified, put a card from your #yExhaust pile into your hand."
+      "The first time a card is #yAmplified each turn, put ",
```
modargo, line 13: Minor, but with how this is set up, the number here won't be highlighted (when the power amount is >1). You'll need to make a separate entry with a #b for the >1 case (since the =1 case doesn't need highlighting).

src/main/resources/anniv5Resources/localization/eng/cosmoscommandpack/Cardstrings.json
```diff
@@ -21,7 +21,7 @@
   },
   "${ModID}:Recall": {
     "NAME": "Recall",
-    "DESCRIPTION": "Whenever a card is ${ModID}:Amplified, put a card from your Exhaust pile into your hand."
+    "DESCRIPTION": "The first time a card is ${ModID}:Amplified each turn, put a card from your Exhaust pile into your hand."
```
modargo, line 24: Would it make more sense to have this stack by affecting the first X cards that Amplify each turn? That's how the base game effects for "the first/next time" stack. I'm mostly thinking of Echo Form and other duplication effects (you don't play the first card X times, you play the first card X cards twice), but this is also true for things like the double damage effect from Phantasmal Killer (you don't get triple damage, you get double damage for more turns).

You can probably find a counter-example somewhere, but in most cases where increasing the number of cards affected or extending the duration makes sense, that's what stacking a power does.

modargo, line 24: Another thought I had given how abusable this power is -- have you considered adding onto this "It gains Purge."?

This mechanic is very abusable since there are some exhausting cards that heal or give you permanent benefits like upgrades (Knighthood) or potions (Alclawmize). Getting those back once is cool. Getting those back an unlimited number of times is abusable. That's why I'm suggesting this as a way to keep the fun and power of this mechanic, but prevent it from being abused by bringing back the same card over and over.

ActuallyEnbeon: As in, the card that was Amplified is Purged? That could be a really cool effect!
Like, you have to sacrifice one card to get another back... I like it

ActuallyEnbeon: Or do you mean the card that was Exhumed... either way, great idea ^^

ActuallyEnbeon: That sounds sensible. Question though: do you think the power should "activate" and count up even if the Exhaust pile is empty?

modargo: I was thinking the card that was Exhumed. I came up with that because it was the most direct way to prevent abusive interactions, and becuase it makes you have to think a bit more -- "do I get back my best card from Exhaust on this turn, or do I wait until next turn when I might need it more, since I can only get it back once?" Otherwise, if you have something like an Adrenaline (or any other generically very strong card) in the Exhaust pile, you'll just get that back turn after turn after turn.

Having the card that was Amplified purge is another interesting idea. It imposes a different kind of limit, since you'll eventually run out of Amplify cards. Though I still like the idea of having the Exhumed card Purge the most since I think it makes for more strategic and interesting gameplay while also preventing degenerate interactions.

modargo: Good question. It seems to me that the case where your exhaust pile is empty is uncommon, and this would only affect the case where you have this power applied, you've exhausted nothing yet (even though you are probably drafting cards to take advantage of this), and you have both an exhaust card and amplify card in hand.

To me, that says that the best thing to do is to keep this simple -- extra complexity for a relatively niche case doesn't seem like its worth it.

If you did want to skip the activation when the exhaust pile was empty, IMO it should be stated on the card/power -- e.g., "The first time a card is Amplified each turn while your exhaust pile isn't empty, ..."

ActuallyEnbeon: Nice, started work on the "card that was Exhumed Purges" version

modargo: Great. I'll take another pass through the review once you've updated it.

For making the exhumed card purge, I'd recommend a card modifier. I don't know if a `PurgeMod` already exists, but you should be able to adapt something like `ExhaustMod`.

ActuallyEnbeon:
Alrighty, made those changes. I also spotted a bug with Cognition so I thought I'd get that in too

modargo:
Looks good to me!

PR #398, opened 2023-01-31
The basics pack is back!

GameboyJ:
My second revision of the basics pack. I hope it now seems fine, if not i'll change cards again, just let me know.
Includes a Hat and card background


mikemayhemdev:
General pack review:

Basic Wave - I like this as the core in which everything else is built around.  The declaration of the pack's intent.  It is an iron wave, but as long as there are sufficient synergies, this should be fine.  The problem is that is doesn't inherit any basic scaling effects that only hit existing cards, such as these multi-upgrade ones.  I might resolve that very directly by doing something like "Counts as a Basic.", and upgrading it upgrades the Strikes/Defends played.

Lose It - I also like this as a general card.  It speaks to what this pack will be about.

Familiar Kit - Wordy, but also a very solid choice to loudly proclaim the pack's identity.  I like the Pickup effects here and wouldn't mind seeing one for Strike and one for Defend, personally.

-----

Simplified Strike - It's very hard in this current iteration to add more basics or generate them.  So this doesn't feel terribly comparable to Perfected Strike.  The only way to make this better is to draft Familiars Kits, Lose Its, or to mirror-dupe basic cards.  It may no longer be a useful pick since moving away from a lot of token generation.

Simplify / Perfect Basics / Basic Supply / Back to Basic - These feel like they are all encroaching each other's design space.  There are four cards dedicated to this draw/upgrade space, and the ultimate result of these is generalized scaling for existing cards only (not generated ones like Basic Wave provides), and that's not to mention Familiar Kit is also here which is providing scaling.   
 This leads to the pack feeling one-dimensional - all it can do is scale up basics and play them.  Some of these slots could be dedicated to other spaces that serve different roles to better flesh out the different purposes and roles of the draft.  Of them, I like Simplify the most as it is doing both the role of scaling and filtering, and is pickable in non-basic-focused drafts.  Compare to Perfect Basics and Basic Supply, which both need to be online to do what Simplify is already doing.  Back to Basics could also probably survive, but the two powers feel like they could be serving other design spaces to make the pack feeling like it has more dimensions to it.

Basic Poison - This feels out of place with the rest - you are already providing scaling win conditions with the multiupgrade path, which also asks you to play lots of basics.  Poison in this case is doing what other cards are already doing.  Any other debuff would be more relevant to making the draft more versatile at doing more jobs, like Weak or Vulnerable.

Occams Razor is definitely too strong as written.  It would be hard to ever make this not broken without a serious restriction.  Something every 3rd basic card may work.

-----

Card Suggestions: An Expunger style X-cost card that generated a Strike with additional modifiers would give this card a heavy hitting finisher that is more versatile than Simplified Strike.  (might be a good place for a Pickup: Obtain a Strike.)

Double Tap style skill, maybe next 2(3) Basics, might play out better than current Occams effect, unless it was nerfed heavily.

Cards that generate big basics, like a 0-cost exhaust card that makes a Defend and upgrades it 2(3) times.  Something that lets you burst ahead to value as you setup in the early turns for these basic amplification effects.  (might be a good place for a Pickup: Obtain a Defend.)

No love for Cardistry?  Would love to see a Lose It style card for Cardistry too, it fits the theme of the pack well.  Maybe could be a Power, something like "Add a Cardistry to hand. The next X times you use Cardistry, gain E.".  An energy spike but in a way that prevents it from going infinite.

GameboyJ:
So, here is what i changed:

Basic Wave is now changes to be a Basic Card when you add it to your deck. Additional upgrades now work correctly.

I removed Basic Poison, Basic Supply and Perfect Basics, and totally redid Simplified Strike and Ockhams Razor

Simplified strike is now a 1 cost Unnate rare attack that deals 3 damage for each Basic card you play this combat. This is a finisher for the playing basic cards build.

Ockhams Razor is now a 1 cost uncommon power that deals 1 damage to all enemies whenever you play a Basic card. It is some chip damage for the playing basic cards build.


Furthermore, I added the following cards:

Conjure Defends. Like conjure blade, except it makes a defend which is played X additional times and is uncommon.

Simple Trick, for 0, adds a Cardisty to your hand and the next 2(3) give you energy.

Strike Plus. costs 1, and adds a Strike+5(7) to your hand.

I listened to some of your suggestions, but most are still changed a bit. I hope you like them, and if not, please tell me again why and i'll make more changes

mikemayhemdev:
Conceptually this is much much better.  Looking forward to trying it out.  The only gut check thing that came to mind while reading was the new Razor maybe being a little weak at 1 damage, since After Image works on everything, but I'd have to play it to see.

GameboyJ:
i think you are mixing up after image and a thousand cuts (which, i'll remind you, costs 2), or you misread razors effect
Thanks for the feedback btw, i am already much happier with this iteration, when i played it, it was fun and some cards were even good when not going for a basics deck now!

mikemayhemdev:
Oh yeah, was definitely in my head thinking of thousand cuts.  The main difference is thousand cuts is tuned for a character that can rapidly generate Shivs.  In an average PM run, even knowing you are running basics pack, I imagine basics are making up at best 50% of your deck, and without tons of token generation.  But, looking through it again, I realize the multi-play nature of Conjure Defends and Lose It might contribute more to Shiv-like gameplay than I'm imagining without playing it.

Alchyr:
I would agree with razor probably being a bit undertuned; thousand cuts synergizes decently well with silent's stuff and is still generally a mediocre pick, especially unupgraded.

Razor always costs 1, which certainly helps, but I think it'd be fine at 2/3 damage per basic due to the limitation.

mikemayhemdev:
General note on the card frame, I would go darker on the tone if possible.  The text is a little hard to read on the light tan backdrop.  Or, use my templater to create a darker section just around the text.

Lose It - Doesn't feel like it needs Innate to be good (and Innate on commons make them very hard to take multiples of), and would make the card able to have better line breaks without breaching 5 lines.

(Obtain a Strike/Defend) - I think my recommendation on the Obtain was wrong.  The new version of the pack, particularly the Simplified Strike addition, makes you not want to obtain more, so I'll admit this was a bad suggestion.  

Conjure Defends - Slight polish note, would love to see the "Played X additional times" added to the Defend preview line.  As for the card itself, regardless of its tuning, it just didn't feel satisfying to play. A card that requires a turn of setup, then to draw later, is hard to become compelling if that bonus when drawn later is also not advancing the combat forward any.  I have some ideas about this, but they all involve a more immediate benefit, like the Strike equivalent does.  I do like the multi-play nature, so my first inclination would be something like add the superdefend to the hand instead of the discard, and add to the end of the card text "Gain E".  So you can always spend the defend you make if you need to use it that turn, or spend on something else if you truly were just needing a setup turn.

Strike Plus - Probably should be Uncommon, it's very strong.  And would change the text to just say "Strike+5", saves a lot of card room.  Also it should pre-upgrade the previewcard to know what a Strike+5 is.

Simple Trick - Should definitely Exhaust, to have parity with Lose It.  Or be a Power, which isn't unheard of for a temporary effect - there are plenty of examples of that in PM, and also Buffer exists.  It's just too much easy value at 0 without exhaust.

Simplified Strike - Surprisingly strong!  Would recommend this cost 2. It isn't hard to get this to be bludgeon+++ levels of damage and the pack is lacking in any high cost cards.  The Unnate mechanic is super strong on this, strong enough to perhaps warrant it being its upgrade.  The text is not quite accurate as well, it reads as a multihit right now and should read more like Blizzard's text wording.  Or do Brilliance style, but it would have to include base damage.

Familiar Kit - It's sad that Zap can't multi-upgrade, and given the bizarreness of the set, I'm tempted to say make an exception rule that says Zap+2 and beyond increases the Lightning generated.  It'd be probably best to do as a custom card rather than the original Zap if so.  It would make for a special 'wow' moment, but understood if you'd prefer not to do this.

Back to Basic - Sadly never saw this in the run, but in practice my draw and energy gain was covered by Simple Trick, cheap Rummage, and Simplify.  I was already capable of doing pretty degenerate things in regards to getting through the deck quickly. What the deck really wanted was as many ways as possible to recur the giant strikes, giant defends, and Simplified Strike.  The other identified issue is that this is a rare that really only synergizes super well with a heavy draft of this same pack, which Simplified Strike is also doing.  Ideally there aren't too many rares that are hard picks when playing with other packs as the primary of the draft.  So, I have a pitch here (actually Vex's pitch who was spectating): "(Retain). 0-cost, Return a discarded card to your hand. If the chosen isn't Basic, Exhaust this."  This would make it pretty pickable to recur any card in any draft, but is especially strong when using it to recur super amped up basic cards.

Occams Razor - Still felt like this wanted to be 2/3, and I was playing almost exclusively this pack.  The pack doesn't really encourage you to play a lot of basics all the time now, to the point I was still actually removing my strikes/defends in shop - the meat of the pack is performing crazy Cardistry/Rummage combos into huge beefed up superstrikes and superdefends and the finisher.

GameboyJ:
okay, the newest iteration is here! Sadly i was asleep while you were playing the run, but i appreciate the feedback. I implemented most things, but in the end left Zap untouched, as it felt weird "changing" a cards upgrades. I like the idea for the new Back to Basic (and it helped me learn more about coding, so thanks for that)

if you think there is anything else which needs changing, let me know!

mikemayhemdev:
Good to go!  Only last little minor polish thing is that, whenever possible, Exhaust should be on a separate line for consistent formatting with base game.  Lose It, Strike Plus, and Simple Trick should follow this rule.

src/main/java/thePackmaster/cardmodifiers/basicspack/DuplicateModifier.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    public void onInitialApplication(AbstractCard card) {
+        super.onInitialApplication(card);
+    }
+
+    @Override
+    public void onUse(AbstractCard card, AbstractCreature target, UseCardAction action) {
+        super.onUse(card, target, action);
```
erasels, line 35: Does the super.onUse have anything you need executed? isn't it just an empty method?

src/main/java/thePackmaster/cardmodifiers/basicspack/DuplicateModifier.java
```diff
@@ -0,0 +1,59 @@
...
+public class DuplicateModifier extends AbstractMadScienceModifier {
+
+    public DuplicateModifier(int valueIn){
+        super(valueIn);
+    }
+
+    @Override
+    public String modifyDescription(String rawDescription, AbstractCard card) {
+        if (this.value == 1) {
+            return rawDescription + CardCrawlGame.languagePack.getUIString(SpireAnniversary5Mod.makeID("DuplicateModifier")).TEXT[0];
```
erasels, line 24: Save the UI string in a static variable instead of retrieving it every frame for every card it's on

src/main/java/thePackmaster/cardmodifiers/basicspack/DuplicateModifier.java
```diff
@@ -0,0 +1,59 @@
...
+        } else return rawDescription + CardCrawlGame.languagePack.getUIString(SpireAnniversary5Mod.makeID("DuplicateModifier")).TEXT[1] + this.value + CardCrawlGame.languagePack.getUIString(SpireAnniversary5Mod.makeID("DuplicateModifier")).TEXT[2];
+    }
+
+    @Override
+    public void onInitialApplication(AbstractCard card) {
+        super.onInitialApplication(card);
```
erasels, line 30: useless

src/main/java/thePackmaster/cardmodifiers/basicspack/DuplicateModifier.java
```diff
@@ -0,0 +1,59 @@
...
+import com.megacrit.cardcrawl.cards.CardQueueItem;
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.core.CardCrawlGame;
+import com.megacrit.cardcrawl.core.Settings;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.monsters.AbstractMonster;
+import thePackmaster.SpireAnniversary5Mod;
+import thePackmaster.cardmodifiers.madsciencepack.AbstractMadScienceModifier;
+
+public class DuplicateModifier extends AbstractMadScienceModifier {
```
erasels, line 15: Very similar to the Expand modifier, if you look at that you'll find this code
`GameActionManager.queueExtraCard(funCard, (AbstractMonster) target);`
Which is a little helper method that you could use

src/main/java/thePackmaster/cards/basicspack/BasicWave.java
```diff
@@ -0,0 +1,74 @@
...
+    public BasicWave() {
+        super(ID, 1, CardType.ATTACK, CardRarity.COMMON, CardTarget.ENEMY, "basics");
+        this.cardToPreview.add(new Strike());
+        this.cardToPreview.add(new Defend());
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        for(int i = 0; i<this.cardToPreview.size(); i++){
+            AbstractCard c = this.cardToPreview.get(i).makeStatEquivalentCopy();
```
erasels, line 36: Again, there's a helper method in GameActionManager for this

src/main/java/thePackmaster/cards/basicspack/ConjureDefends.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class ConjureDefends extends AbstractPackmasterCard {
+    public final static String ID = makeID("ConjureDefends");
+
+    public ConjureDefends() {
+        super(ID, -1, CardType.SKILL, CardRarity.UNCOMMON, CardTarget.SELF, "basics");
+        this.exhaust = true;
+        this.cardsToPreview = new Defend();
+        this.cardsToPreview.upgrade();
+        this.cardsToPreview.rawDescription += CardCrawlGame.languagePack.getUIString(SpireAnniversary5Mod.makeID("DuplicateModifier")).TEXT[1] + "X" + CardCrawlGame.languagePack.getUIString(SpireAnniversary5Mod.makeID("DuplicateModifier")).TEXT[2];
```
erasels, line 32: Is this X fine for translation? Too lazy to check the strings right now

src/main/java/thePackmaster/cards/basicspack/ConjureDefends.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        AbstractCard card = new Defend();
+        if(this.upgraded){
+            addToBot(new ConjureCardsAction(p, this.freeToPlayOnce, this.energyOnUse+1, card));
+        } else addToBot(new ConjureCardsAction(p, this.freeToPlayOnce, this.energyOnUse, card));
+        if(this.energyOnUse>0 && !this.freeToPlayOnce) {
+            addToBot(new GainEnergyAction(1));
```
erasels, line 43: Huh? Refund is an stslib feature iirc

src/main/java/thePackmaster/cards/basicspack/FamiliarKit.java
```diff
@@ -0,0 +1,54 @@
...
+        this.baseSecondMagic = this.secondMagic = 1;
+        this.cardToPreview.add(new Bash());
+        this.cardToPreview.add(new Survivor());
+        this.cardToPreview.add(new Zap());
+    }
+
+    @Override
+    public void onObtainCard() {
+        for(int i = 0; i<this.cardToPreview.size(); i++){
+            float width = Settings.WIDTH / 2.0F - 150 * this.cardToPreview.size() + 300 * i;
```
erasels, line 38: 150 and 300 aren't multipled by scale so different resolutions will have issues

src/main/java/thePackmaster/cards/basicspack/LoseIt.java
```diff
@@ -0,0 +1,51 @@
...
+        this.cardsToPreview.isCostModified = true;
+        this.baseMagicNumber = this.magicNumber = 1;
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        addToTop(new MakeTempCardInHandAction(cardsToPreview));
+        for(AbstractCard c : p.hand.group)
+            if (c.cardID.equals(Rummage.ID) && c.cost > this.magicNumber) {
+                c.costForTurn = c.cost = this.magicNumber;
```
erasels, line 33: This is not for combat

src/main/java/thePackmaster/cards/basicspack/SimplifiedStrike.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        int basicCount = 0;
+        for (AbstractCard c : AbstractDungeon.actionManager.cardsPlayedThisCombat) {
+            if (c.rarity == CardRarity.BASIC)
+                basicCount++;
+        }
+        this.baseDamage = this.magicNumber * basicCount;
```
erasels, line 31: Kinda jank. You could just calculate it in calculateCardDamage correctly so you didn't have to do this.

src/main/java/thePackmaster/cards/basicspack/SimplifiedStrike.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public void applyPowers() {
+        int basicCount = 0;
+        for (AbstractCard c : AbstractDungeon.actionManager.cardsPlayedThisCombat) {
+            if (c.rarity == CardRarity.BASIC)
+                basicCount++;
+        }
+        if (basicCount > 0) {
+            this.baseDamage = this.magicNumber * basicCount;
+            super.applyPowers();
```
erasels, line 50: Not called at all if basic count is <0? Not a fan of that

src/main/java/thePackmaster/cards/basicspack/StrikePlus.java
```diff
@@ -0,0 +1,43 @@
...
+import thePackmaster.cards.AbstractPackmasterCard;
+import thePackmaster.cards.Defend;
+import thePackmaster.cards.Strike;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class StrikePlus extends AbstractPackmasterCard {
+    public final static String ID = makeID("StrikePlus");
+
+    public StrikePlus() {
```
erasels, line 20: Hmm, having skills that are strikes are kinda jank. Would you mind renaming it? if not, at least give it the strike tag for consistency

src/main/java/thePackmaster/cards/basicspack/StrikePlus.java
```diff
@@ -0,0 +1,43 @@
...
+            this.cardsToPreview.upgraded = false;
+            this.cardsToPreview.upgrade();
+        }
+        this.cardsToPreview.name = this.cardsToPreview.originalName + "+" + this.cardsToPreview.timesUpgraded;
+        this.exhaust = true;
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        addToBot(new MakeTempCardInHandAction(this.cardsToPreview));
```
erasels, line 33: Is the given card correctly upgraded? Not sure if makeStatEquivCopy actually upgrades beyond the base upgrade

src/main/java/thePackmaster/powers/basicspack/SimpleTrickPower.java
```diff
@@ -0,0 +1,45 @@
...
+    public SimpleTrickPower(int amount) {
+        super(POWER_ID, NAME, PowerType.BUFF, false, AbstractDungeon.player, amount);
+    }
+
+    @Override
+    public void onUseCard(AbstractCard card, UseCardAction action){
+        if (card.cardID.equals(Cardistry.ID) ) {
+            flash();
+            addToBot(new GainEnergyAction(1));
+            this.amount--;
```
erasels, line 28: powers have a method called reducePower :)

src/main/resources/anniv5Resources/localization/eng/basicspack/Cardstrings.json
```diff
@@ -0,0 +1,60 @@
...
+    ]
+  },
+  "${ModID}:FamiliarKit": {
+    "NAME": "Familiar Kit",
+    "DESCRIPTION": "Gain !M! Strength, !M! Dexterity and !${ModID}:m2! Focus. ${ModID}:Pickup: Obtain a *Bash, *Survivor and *Zap.",
+    "UPGRADE_DESCRIPTION": "Innate. NL Gain !M! Strength, !M! Dexterity and !${ModID}:m2! Focus. ${ModID}:Pickup: Obtain a *Bash, *Survivor and *Zap."
+  },
+  "${ModID}:OckhamsRazor": {
+    "NAME": "Ockhams Razor",
+    "DESCRIPTION": "Whenever you play a ${ModID}:Basic card, deal !M! damage to all enemies."
```
erasels, line 36: ALL

src/main/resources/anniv5Resources/localization/eng/basicspack/Cardstrings.json
```diff
@@ -0,0 +1,60 @@
...
+  },
+  "${ModID}:BackToBasic": {
+    "NAME": "Back To Basic",
+    "DESCRIPTION": "Put a card from your discard pile into your hand. If it was not a ${ModID}:Basic card, Exhaust.",
+    "UPGRADE_DESCRIPTION": "Retain. NL Put a card from your discard pile into your hand. If it was not a ${ModID}:Basic card, Exhaust."
+  },
+  "${ModID}:StrikePlus": {
+    "NAME": "Strike Plus",
+    "DESCRIPTION": "Add a *Strike+5 to your hand. NL Exhaust.",
```
erasels, line 45: Could use the `Gain a CardName.` syntax we've used a lot in packmaster to say Put a copy of CardName into your hand.

Could also use magicNumber for the strike upgrade amount

src/main/resources/anniv5Resources/localization/eng/basicspack/UIstrings.json
```diff
@@ -0,0 +1,22 @@
...
+  "${ModID}:BasicsPack": {
+    "TEXT": [
+      "Basics",
+      "A pack centered around very Basic cards.",
+      "Jack Renoson"
+    ]
+  },
+  "${ModID}:BasicsPackHat": {
+    "TEXT": [
+      "A Basic hat for a Simple adventurer.",
```
erasels, line 18: Why is simple capitalized?

src/main/resources/anniv5Resources/localization/eng/basicspack/Powerstrings.json
```diff
@@ -0,0 +1,17 @@
...
+    "DESCRIPTIONS": [
+      "Whenever you play a #yBasic card, deal ",
+      " damage to all enemies"
+    ]
+  },
+  "${ModID}:SimpleTrickPower": {
+    "NAME": "Simple Trick",
+    "DESCRIPTIONS": [
+      "The next time you play a #yCardistry, gain [E] .",
+      "The next ",
```
erasels, line 13:
```suggestion
      "The next #b",
```

erasels (changes requested):

src/main/java/thePackmaster/cards/basicspack/ConjureDefends.java
GameboyJ: The X is so that the preview says Played X times. If all languages use X for X-cost cards, it is fine...

PR #400, opened 2023-01-31
Scry plus

Jedi515:


erasels:
I honestly think this card has too much scry. Maybe try going for some implicit synergy like putting cards back or an effect like rummage that draws a card and modifies it or Escape Plan.

Right now this pack seems very surface level and not very fun to play due to the excess of not very useful scry.

Mwallx:
Some instance idea as a negative effect for Scry to balance out things: Reording your draw pile

PR #407, opened 2023-02-01
Farmer Pack minor rework and clarity

NoMasPollo:


src/main/resources/anniv5Resources/localization/eng/farmerpack/Cardstrings.json
```diff
@@ -1,8 +1,13 @@
 {
   "${ModID}:Fertilizer": {
     "NAME": "Fertilizer",
-    "DESCRIPTION": "Draw a card. NL Exhaust.",
-    "UPGRADE_DESCRIPTION": "Draw !M! cards. NL Exhaust."
+    "DESCRIPTION": "Draw a card. NL Exhaust. NL This is a Status",
```
erasels, line 4: Is the status text really needed? that's what the type is for.
Also, would need a period

erasels:
Please also update the zhs counterpart of your strings (just overwrite them with the english text)

NoMasPollo:
Changes to ZHS made, edited the text that changed. Also, I do think that it is pertinent to put "This is a Status" because almost everyone I've seen play the pack either don't realize its a status or take a significant amount of time. Not enough people check the type line for it to be noticeable, apparently.

PR #410, opened 2023-02-03
feat: Rippable Card Rework

EricBartusch:
Fundamentally changed how Rippable cards work in the code.
Removed AbstractRippableCard/AbstractArtCard/AbstractTextCard. All rippable cards are now just the original card with shaders applied and behave differently. They no longer utilize card descriptors, but change the title of the card.

Created RippableCardModifier to signify when cards can be ripped.
Added an interface for adding on rip effects.
Added an action to rip cards in hand.
You can now rip cards even if it'll result in card halves being put into the discard pile.

Changed Divided Fury to now make a random card in hand rippable at the start of the turn.

Improved Smithing Hammer's VFX


src/main/resources/anniv5Resources/localization/eng/rippack/Cardstrings.json
```diff
@@ -1,8 +1,8 @@
 {
   "${ModID}:DividedFury": {
     "NAME": "Divided Fury",
-    "DESCRIPTION": "At the start of your turn, add a *Fury into your hand.",
-    "UPGRADE_DESCRIPTION": "Innate. NL At the start of your turn, add a *Fury into your hand."
+    "DESCRIPTION": "At the start of your turn, make !M! card in your ${ModID}:rippable.",
+    "UPGRADE_DESCRIPTION": "At the start of your turn, make !M! card in your ${ModID}:rippable."
```
erasels, line 5:
```suggestion
    "UPGRADE_DESCRIPTION": "At the start of your turn, make !M! cards in your ${ModID}:rippable."
```

erasels (changes requested):
Please also update the zhs counterpart of your strings (just overwrite them with the english text)

src/main/resources/anniv5Resources/localization/eng/rippack/Cardstrings.json
EricBartusch: !M! is always one

erasels (approved):

PR #416, opened 2023-02-05
Fix pack filtering and hat menus being positioned incorrectly at resolutions that aren't a 16:9 ratio

modargo:
Not sure why this wasn't fixed yet, but I assume it must be because people aren't aware that Settings.xScale and Settings.yScale exist and are different from Settings.scale. Those are what you need to use if you want things to work at ratios other than the game's base 16:9... but even the buttons provided by BaseMod don't use them. I bet there are mods out there with weird looking config panels on non-standard resolution ratios (the Packmaster config menu had things positioned wrong because of this).

To show that this works, here's how it looks at 1280x1024 (a resolution that's "taller" than normal), at 1680x720 (a resolution that's "wider" than normal), and 1280x720 (a resolution that's 16:9 but smaller than the base 1920x1080):

image
image
image

As you can see, some UI elements get stretched in slightly odd ways, and the UI isn't smart enough to compensate for the character image staying the same size while other stuff changes, but there's no major issues. These three test resolutions cover a pretty broad set of cases (wide/tall/normal), so this is probably enough to solve the issues people have been complaining about.

Leaving this one up for you to review since it's worth spreading the word about Settings.xScale and Settings.yScale and making sure any future UI code uses them, particularly when it comes to positioning. There are valid uses of Settings.scale (mostly for having a UI element scale up or down but keep its width-to-height ratio the same regardless of resolution), but positioning usually needs xScale/yScale to be correct.

### DIFF
--- a/src/main/java/thePackmaster/SpireAnniversary5Mod.java
+++ b/src/main/java/thePackmaster/SpireAnniversary5Mod.java
@@ -2,7 +2,6 @@ package thePackmaster;
 
 import basemod.AutoAdd;
 import basemod.BaseMod;
-import basemod.ModLabeledToggleButton;
 import basemod.ModPanel;
 import basemod.abstracts.CustomSavable;
 import basemod.devcommands.ConsoleCommand;
@@ -91,6 +90,7 @@ import thePackmaster.stances.downfallpack.AncientStance;
 import thePackmaster.stances.sentinelpack.Angry;
 import thePackmaster.stances.sentinelpack.Serene;
 import thePackmaster.ui.CurrentRunCardsTopPanelItem;
+import thePackmaster.ui.FixedModLabeledToggleButton.FixedModLabeledToggleButton;
 import thePackmaster.ui.InfestTextIcon;
 import thePackmaster.ui.PackFilterMenu;
 import thePackmaster.util.JediUtil;
@@ -1047,7 +1047,7 @@ public class SpireAnniversary5Mod implements
         settingsPanel = new ModPanel();
         //int configStep = 40;
 
-        ModLabeledToggleButton allPacksModeBtn = new ModLabeledToggleButton(configStrings.TEXT[3], 350.0f, 600F, Settings.CREAM_COLOR, FontHelper.charDescFont, allPacksMode, settingsPanel, (label) -> {
+        FixedModLabeledToggleButton allPacksModeBtn = new FixedModLabeledToggleButton(configStrings.TEXT[3], 350.0f, 600F, Settings.CREAM_COLOR, FontHelper.charDescFont, allPacksMode, settingsPanel, (label) -> {
 
         }, (button) -> {
             allPacksMode = button.enabled;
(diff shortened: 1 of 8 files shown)


src/main/java/thePackmaster/ui/FixedModLabeledToggleButton/FixedModLabel.java
```diff
@@ -0,0 +1,90 @@
...
+import basemod.ModPanel;
+import com.badlogic.gdx.graphics.Color;
+import com.badlogic.gdx.graphics.g2d.BitmapFont;
+import com.badlogic.gdx.graphics.g2d.SpriteBatch;
+import com.megacrit.cardcrawl.core.Settings;
+import com.megacrit.cardcrawl.helpers.FontHelper;
+
+import java.util.function.Consumer;
+
+// This class is a near-copy of the BaseMod class, changed to use Settings.xScale and Settings.yScale
```
modargo, line 13: In theory, it would be best if the versions of these classes in BaseMod were replaced with what I made here. (The code for those classes can be found at https://github.com/daviscook477/BaseMod/tree/master/mod/src/main/java/basemod).

In practice, mods using the BaseMod classes that rely on or work around the current functionality (i.e. using only `Settings.scale`, not `Settings.xScale` and `Settings.yScale`) could easily be broken by replacing these, and dealing with that beyond what I think is wise to take on when the immediate goal is just to fix this issue for Packmaster.

src/main/java/thePackmaster/hats/HatMenu.java
```diff
@@ -53,25 +53,26 @@ public class HatMenu {
     private static final TextureRegion MENU_BG = new TextureRegion(ImageMaster.loadImage("img/ModPanelBg.png"));
 
     //positions
-    private static final float BG_X_SCALE = Settings.scale * 0.275f;
-    private static final float BG_Y_SCALE = Settings.scale * 0.8f;
-    private static final float BG_X = 525f * Settings.scale;
-    private static final float BG_Y = Settings.HEIGHT - 40f * Settings.scale - MENU_BG.getRegionHeight() * BG_Y_SCALE;
+    private static final float BG_X_SCALE = Settings.xScale * 0.275f;
```
modargo, line 56: Most of the changes are just like this -- take something using `Settings.scale`, switch it to using whichever of `Settings.xScale` or `Settings.yScale` is appropriate, problem solved.

src/main/java/thePackmaster/SpireAnniversary5Mod.java
```diff
@@ -1047,7 +1047,7 @@ private void initializeConfig() {
         settingsPanel = new ModPanel();
         //int configStep = 40;
 
-        ModLabeledToggleButton allPacksModeBtn = new ModLabeledToggleButton(configStrings.TEXT[3], 350.0f, 600F, Settings.CREAM_COLOR, FontHelper.charDescFont, allPacksMode, settingsPanel, (label) -> {
+        FixedModLabeledToggleButton allPacksModeBtn = new FixedModLabeledToggleButton(configStrings.TEXT[3], 350.0f, 600F, Settings.CREAM_COLOR, FontHelper.charDescFont, allPacksMode, settingsPanel, (label) -> {
```
modargo, line 1050: The changes in this file are for the config menu -- a secondary issue that I only noticed because I looked at where else this button was used.

PR #421, opened 2023-02-06
Biting Cold rebalancing & bugfix

ActuallyEnbeon:
Made the basic Frostbite application cards stronger and also fixed Growing Affliction entering hand when enemies applied debuffs
Also fixed a tiny thing in Frostbite's power strings


ActuallyEnbeon:
I might rework Growing Affliction at some point to mitigate the infinite potential but also it's kind of hard to do and pretty fun

PR #425, opened 2023-02-08
Relic stats, run history, and a few bug fixes

modargo:
This PR has a few different changes. I started doing one thing (relic stats), then wanted to play a few runs to test it since the code to get this working is sometimes a bit quirky, then kept finding new QOL stuff to add. There were a variety of things (mostly in run history) that I realized wanted to add for the Packmaster.

Here's what you can include in the changelog for this:
* Add relic stats for Bag of Holding, Banishing Decree, Collector Badge, and Gem of Unity, PM Booster Box, PM Booster Pack, and PM Collection
* Show the list of packs in run history for Packmaster runs
* Replicators pack: fix Energy Drink being marked as a turn-based buff
* Character event: add text coloring for remove card option
* Character event: show event name in run history
* Character event: show event choices in run history
* Character event: fix the relic options not costing gold

It shouldn't be too hard to tell which code changes are associated with each changelog entry. The files involved are different for each one and the code is mostly straightforward, which is why I didn't think it was worth the time to split this into multiple different PRs.

### DIFF
--- a/src/main/java/thePackmaster/events/BlackMarketDealerEvent.java
+++ b/src/main/java/thePackmaster/events/BlackMarketDealerEvent.java
@@ -171,12 +171,14 @@ public class BlackMarketDealerEvent extends PhasedEvent {
 
 
                 .addOption(new TextPhase.OptionInfo(hasCollection() ? OPTIONS[23] : hasEnoughGoldForBuyCollection() ? OPTIONS[15] + getGoldCostForBuyCollection() + OPTIONS[16] : OPTIONS[9] + getGoldCostForBuyCollection() + OPTIONS[7], new PMCollection()).enabledCondition(this::canBuyCollection), (i) -> {   //PM Collection
+                    AbstractDungeon.player.loseGold(getGoldCostForBuyCollection());
                     AbstractDungeon.getCurrRoom().spawnRelicAndObtain((float) Settings.WIDTH / 2.0F, (float) Settings.HEIGHT / 2.0F, new PMCollection());
                     AbstractDungeon.shopRelicPool.remove(PMCollection.ID);
                     transitionKey("relicCollectionEnd");
                 })
 
                 .addOption(new TextPhase.OptionInfo(hasDecree() ? OPTIONS[23] : hasEnoughGoldForBuyDecree() ? OPTIONS[17] + getGoldCostForBuyDecree() + OPTIONS[18] : OPTIONS[9] + getGoldCostForBuyDecree() + OPTIONS[7], new BanishingDecree()).enabledCondition(this::canBuyDecree), (i) -> {   //Banishing Decree
+                    AbstractDungeon.player.loseGold(getGoldCostForBuyDecree());
                     AbstractDungeon.getCurrRoom().spawnRelicAndObtain((float) Settings.WIDTH / 2.0F, (float) Settings.HEIGHT / 2.0F, new BanishingDecree());
                     AbstractDungeon.shopRelicPool.remove(BanishingDecree.ID);
                     transitionKey("relicBanishingEnd");
--- a/src/main/java/thePackmaster/events/BlackMarketDealerEvent.java
+++ b/src/main/java/thePackmaster/events/BlackMarketDealerEvent.java
@@ -5,13 +5,13 @@ import basemod.abstracts.events.phases.TextPhase;
 import com.megacrit.cardcrawl.cards.AbstractCard;
 import com.megacrit.cardcrawl.cards.CardGroup;
 import com.megacrit.cardcrawl.cards.curses.CurseOfTheBell;
-import com.megacrit.cardcrawl.cards.curses.Necronomicurse;
 import com.megacrit.cardcrawl.core.CardCrawlGame;
 import com.megacrit.cardcrawl.core.Settings;
 import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
 import com.megacrit.cardcrawl.events.city.Beggar;
 import com.megacrit.cardcrawl.events.city.TheLibrary;
 import com.megacrit.cardcrawl.localization.EventStrings;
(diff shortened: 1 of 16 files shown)


src/main/java/thePackmaster/patches/DisableRelicStatsForChinesePatch.java
```diff
@@ -0,0 +1,20 @@
+package thePackmaster.patches;
+
+import com.evacipated.cardcrawl.modthespire.lib.SpirePatch;
+import com.evacipated.cardcrawl.modthespire.lib.SpirePrefixPatch;
+import com.evacipated.cardcrawl.modthespire.lib.SpireReturn;
+import com.megacrit.cardcrawl.core.Settings;
+import thePackmaster.SpireAnniversary5Mod;
+
+// This exists because relic stats were added after the mod was translated to Chinese (ZHS), so the entries in the
```
modargo, line 9: It's probably worth adding support for this, but I wanted to have this PR entirely functional without needing to coordinate ZHS entries. 

Once this is merged we can see if mwalls is up for adding support for relic stats in ZHS or not. It's mostly just working out where to add the new strings.

src/main/java/thePackmaster/patches/GetEventNamePatch.java
```diff
@@ -0,0 +1,23 @@
...
+import com.megacrit.cardcrawl.core.CardCrawlGame;
+import com.megacrit.cardcrawl.helpers.EventHelper;
+import thePackmaster.SpireAnniversary5Mod;
+
+@SpirePatch(
+        clz = EventHelper.class,
+        method = "getEventName",
+        paramtypez = String.class
+)
+public class GetEventNamePatch {
```
modargo, line 15: A patch like this is needed for every mod that wants custom events to show up properly in run history. Now that I think about it, some form of this should probably be in BaseMod, but it's an easy enough patch that nobody has bothered.

src/main/java/thePackmaster/relics/BagOfHolding.java
```diff
@@ -81,5 +90,49 @@ public String getUpdatedDescription() {
         return DESCRIPTIONS[0] + sb + DESCRIPTIONS[1];
     }
 
+    private static final Map<String, Integer> stats = new HashMap<>();
+    public static String ENERGY_STAT = "energy";
```
modargo, line 94: There's no dependency needed for any of this because the relic stats code looks for these methods by reflection. See https://github.com/ForgottenArbiter/StsRelicStats/blob/184a8584ece3b2feea550d653c16e63976c2c3fd/src/main/java/relicstats/patches/BaseModPatch.java#L28

src/main/java/thePackmaster/relics/BagOfHolding.java
```diff
@@ -81,5 +90,49 @@ public String getUpdatedDescription() {
...
+        if (jsonElement != null) {
+            JsonArray jsonArray = jsonElement.getAsJsonArray();
+            stats.put(ENERGY_STAT, jsonArray.get(0).getAsInt());
+        } else {
+            resetStats();
+        }
+    }
+
+    public static void incrementEnergyStat() {
+        stats.put(ENERGY_STAT, stats.getOrDefault(ENERGY_STAT, 0) + 1);
```
modargo, line 135: This is all written so that even if relic stats isn't loading, everything will be fine -- thus the use of getOrDefault here and elsewhere.

src/main/java/thePackmaster/patches/PacksRunHistoryPatch.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.lang.reflect.Field;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class PacksRunHistoryPatch {
+    private static final String[] TEXT = CardCrawlGame.languagePack.getUIString(SpireAnniversary5Mod.makeID("RunHistory")).TEXT;
+
+    @SpirePatch(clz = CardCrawlGame.class, method = SpirePatch.CONSTRUCTOR)
+    public static class PacksField {
```
modargo, line 26: This is the same basic structure that Run History Plus follows. Add a new field, gather data for it, let it get automatically loaded by JSON magic, and patch into the rendering code to display it.

src/main/java/thePackmaster/events/BlackMarketDealerEvent.java
```diff
@@ -144,21 +152,23 @@ public void update() {
...
                             AbstractDungeon.effectList.add(new ShowCardAndObtainEffect(new ConjurePack(), (Settings.WIDTH * .66F), (float) (Settings.HEIGHT / 2)));
+                            logMetricObtainCards(ID, "Conjure Pack", Arrays.asList(ConjurePack.ID, CurseOfTheBell.ID));
                             transitionKey("magicLearnEnd");
                         }).addOption(OPTIONS[12], (i) -> {   //Pack in a Jar Potion
                             AbstractDungeon.getCurrRoom().rewards.clear();
                             AbstractDungeon.getCurrRoom().rewards.add(new RewardItem(new PackInAJar()));
                             skipDefaultCardRewards = true;
                             AbstractDungeon.combatRewardScreen.open();
                             skipDefaultCardRewards = false;
+                            logMetric(ID, "Pack in a Jar", null, null, null, null, null, Collections.singletonList(PackInAJar.POTION_ID), null, 0, 0, 0, 0, 0, 0);
```
modargo, line 163: There are a few oddities with the event choices like this one on the run history page -- the base game isn't set up to properly display event choices when both gain a relic and gain cards/potions, and the patch to fix that is tricky enough that I decided against doing it right now. You can still work out what happened from what run history displays, it's just missing some information.

modargo:
The only general Packmaster relic without stats is Handy Haversack -- not really anything meaningful to track.

The pack-specific relics could have stats added, but it wasn't something I was interested in doing.

PR #447, opened 2023-02-18
Core Set Updates

mikemayhemdev:
Synergize: Now uses the new "Synergy" keyword, which is the same as the card's previous condition, just wrapped up into a keyword instead of three lines of text.

Backpack Smack: Reworked to be an X-cost card.  Deals 6 damage per energy spent.  Has Synergy: Gain 1 Energy.

Showoff: Block reduced by 1.

Ripped Pecs: Reworked to now gain 1 Strength, plus another 2(3) if you have Synergy.

Smithing Hammer: Damage reduced to 3x2, upgrading to 3 hits on upgrade, but always upgrades 2 random cards instead of 1(2).

### DIFF
--- a/src/main/resources/anniv5Resources/localization/eng/coresetpack/Cardstrings.json
+++ b/src/main/resources/anniv5Resources/localization/eng/coresetpack/Cardstrings.json
@@ -1,7 +1,7 @@
 {
   "${ModID}:BackpackSmack": {
     "NAME": "Backpack Smack",
-    "DESCRIPTION": "Deal !D! damage. NL If your draw pile has 10 or more cards, draw !M! cards. NL Otherwise, gain [E] ."
+    "DESCRIPTION": "Deal !D! damage X times. NL ${ModID}:Synergy: Gain [E] ."
   },
   "${ModID}:BoosterTutor": {
     "NAME": "Booster Tutor",
@@ -9,7 +9,7 @@
   },
   "${ModID}:DestinyDraw": {
     "NAME": "Destiny Draw",
-    "DESCRIPTION": "Choose 1 of 3 random cards from different Packs to add to your hand. NL It costs 0 this turn. NL Exhaust."
+    "DESCRIPTION": "Choose 1 of 3 cards from different Packs to add to your hand. It costs 0 this turn. NL Exhaust."
   },
   "${ModID}:Flick": {
     "NAME": "Flick",
@@ -22,9 +22,7 @@
   },
   "${ModID}:RippedPecs": {
     "NAME": "Ripped Pecs",
-    "DESCRIPTION": "Gain 1 Strength for each Pack you've played a card from this combat. NL Exhaust.",
-    "UPGRADE_DESCRIPTION": "Gain 1 Strength for each Pack you've played a card from this combat.",
-    "EXTENDED_DESCRIPTION": [" NL (Gains {0} Strength.)"]
+    "DESCRIPTION": "Gain 1 Strength. NL ${ModID}:Synergy: Gain !M! Strength."
   },
   "${ModID}:Showoff": {
     "NAME": "Showoff",
@@ -36,11 +34,10 @@
   },
   "${ModID}:SmithingHammer": {
     "NAME": "Smithing Hammer",
-    "DESCRIPTION": "Deal !D! damage 3 times. NL Upgrade a random card in your hand for the rest of combat.",
-    "UPGRADE_DESCRIPTION": "Deal !D! damage 3 times. NL Upgrade !M! random cards in your hand for the rest of combat."
+    "DESCRIPTION": "Deal !D! damage !M! times. NL Upgrade 2 random cards in your hand."
   },
   "${ModID}:Synergize": {
     "NAME": "Synergize",
-    "DESCRIPTION": "Deal !D! damage to ALL enemies. NL If you have cards from at least two other Packs in your hand, gain !B! Block."
+    "DESCRIPTION": "Deal !D! damage to ALL enemies. NL ${ModID}:Synergy: Gain !B! Block."
   }
 }
--- a/src/main/resources/anniv5Resources/localization/eng/coresetpack/Cardstrings.json
+++ b/src/main/resources/anniv5Resources/localization/eng/coresetpack/Cardstrings.json
@@ -9,7 +9,7 @@
   },
   "${ModID}:DestinyDraw": {
     "NAME": "Destiny Draw",
-    "DESCRIPTION": "Choose 1 of 3 cards from different Packs to add to your hand. It costs 0 this turn. NL Exhaust."
+    "DESCRIPTION": "Choose 1 of 3 random cards from different Packs to add to your hand. It costs 0 this turn. Exhaust."
   },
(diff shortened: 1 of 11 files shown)


src/main/resources/anniv5Resources/localization/eng/coresetpack/Cardstrings.json
```diff
@@ -22,9 +22,7 @@
   },
   "${ModID}:RippedPecs": {
     "NAME": "Ripped Pecs",
-    "DESCRIPTION": "Gain 1 Strength for each Pack you've played a card from this combat. NL Exhaust.",
-    "UPGRADE_DESCRIPTION": "Gain 1 Strength for each Pack you've played a card from this combat.",
-    "EXTENDED_DESCRIPTION": [" NL (Gains {0} Strength.)"]
+    "DESCRIPTION": "Gain 1 Strength. NL ${ModID}:Synergy: Gain !M! Strength."
```
erasels, line 25: Having this be 2 power applications feels a bit jank. Might be nicer to say Gain !M! Strength instead and change the use method accordingly.

src/main/resources/anniv5Resources/localization/eng/coresetpack/Keywordstrings.json
```diff
@@ -1,2 +1,9 @@
 [
+  {
+    "PROPER_NAME": "Synergy",
+    "NAMES": [
+      "synergy"
+    ],
+    "DESCRIPTION": "Additional effect if you have cards from 3 different packs in your hand."
```
erasels, line 7: different is a little confusing here. Different would make me think it excludes itself which it doesn't do.

erasels (changes requested):

src/main/resources/anniv5Resources/localization/eng/coresetpack/Cardstrings.json
mikemayhemdev: I'm confused, do you mean that it should not have the 1 strength by default?  or that the use code should total the two, and the text here is correct?

src/main/resources/anniv5Resources/localization/eng/coresetpack/Keywordstrings.json
mikemayhemdev: would 'unique packs' alleviate this?

mikemayhemdev:
Changes submitted.

src/main/resources/anniv5Resources/localization/eng/coresetpack/Cardstrings.json
```diff
@@ -1,15 +1,15 @@
...
-    "DESCRIPTION": "Deal !D! damage. NL If your draw pile has 10 or more cards, draw !M! cards. NL Otherwise, gain [E] ."
+    "DESCRIPTION": "Deal !D! damage X times. NL ${ModID}:Synergy: Gain [E] ."
   },
   "${ModID}:BoosterTutor": {
     "NAME": "Booster Tutor",
     "DESCRIPTION": "Put !M! random cards from different Packs from your draw pile into your hand."
   },
   "${ModID}:DestinyDraw": {
     "NAME": "Destiny Draw",
-    "DESCRIPTION": "Choose 1 of 3 random cards from different Packs to add to your hand. NL It costs 0 this turn. NL Exhaust."
```
modargo, line 12: Why remove the word random here? Discovery and all other cards like it use "random" and I think it's an important word to keep in the text. It helps establish this as a parallel effect to Discovery that will be random just like it.

I'm fine with removing the NL (I see that Discovery doesn't have one), but I think "random" should stay.

mikemayhemdev: 'random' is the difference between 5 lines and 6 lines in this case, that was the reasoning.  Perhaps we can find another shortening elsewhere?  Or leave it as is, but it really doesn't look good at 6 lines.

src/main/resources/anniv5Resources/localization/eng/coresetpack/Cardstrings.json
modargo: Not sure if this was Gk's intent with the comment, but what came to mind for me was the version of this card that was just: `Synergy: Gain 3 (4) Strength`.

It's a bit weaker than `Gain 1 Strength. Gain 2 (3) Strength` version, but it should still be a strong card and I like it for a few reasons: (1) very simple text (2) uses the design space of a do-nothing-without-synergy card (3) pleasing symmetry with Spot Weakness.

mikemayhemdev: If we were to do that, I'd prefer it be "Requires Synergy".  Having a playable card as potent as that which will either be big, or do nothing, depending on if its glowing or not, is a pitfall for players to fall into.  Someone suggested it always gaining 1 Strength, which seemed more approachable.  My preferred would still remain 'Requires Synergy'.

mikemayhemdev: While the same argument can be made of Spot Weakness, you at least have to drag spot weakness onto a target, which has its intent directly listed over it.  It's harder to whiff a spot weakness than to whiff a self-targeted card that has a hand-based conditional.

src/main/resources/anniv5Resources/localization/eng/coresetpack/Cardstrings.json
modargo: It's my point of view that sticking with consistent wording is more important than avoiding 6 lines. It would be very strange to have this be the one card like this that doesn't say "random" (thus implying that its doing something different than everything else, even though it isn't). Especially given that the card in the current released version is already 6 lines.

If there's a way to fix it being six lines with breaking wording consistently, I'm all for it. Otherwise, I think keeping the status quo is better than fixing one problem only to introduce another one.

modargo: On the topic of fixing it being six lines while keeping "random" -- what happens if you remove the NL before Exhaust? I believe there are examples of that approach in other mods (and maybe even somewhere in the base game?). And I like a minor formatting inconsistency like that a lot more than wording inconsistency.

mikemayhemdev: Was already trying that out, and it just baaaarely works.  So shall it be.

src/main/resources/anniv5Resources/localization/eng/coresetpack/Cardstrings.json
modargo: I would be completely fine with the Requires Synergy version.

I don't really feel that it's necessary to have Requires Synergy instead of just Synergy, but I understand how you can draw a distinction between this and Spot Weakness. And even with the extra word, I still like the Requires Synergy version more than the 1 Strength/Synergy: 2 (3) Strength version.

Really not something I feel that strongly about, but there's something in me that is pulled towards the version of this card that does nothing without Synergy. All the stuff I said in my first comment still applies, and it just feels better somehow.

PR #458, opened 2023-02-24
Paint brush nerf and rarity fixes

Jedi515:


erasels:
Removed 2 cards from uncommon rarity, is this intended?

PR #460, opened 2023-02-24
Serpentine pack: change the Tainted debuff from Sinister Concoction so that additional stacks provide the same increase as the first stack (+100% per stack instead of +200%)

modargo:
Also clean up dead code for an old version of the debuff.

Leaving this up so that you can for you do a quick review of the changes and note this for the changelog.


src/main/java/thePackmaster/powers/serpentinepack/TaintedEnemy.java
```diff
@@ -25,18 +25,16 @@ public TaintedEnemy(AbstractCreature owner, int amount) {
     public void updateDescription() {
         if (this.amount == 1) {
             this.description = DESCRIPTIONS[0];
-        } else if (Settings.language == Settings.GameLanguage.ZHS) {
-            this.description = DESCRIPTIONS[1] + this.amount*2 + DESCRIPTIONS[2];
```
modargo, line 29: In the old code, `this.amount*2` was the factor that debuffs were multiplied by. In the new code, it's `this.amount+1`.

Among other things, this means that the ZHS description for 2+ stacks was already using the actual multiplication factor (instead of the number of stacks). With my changes, both languages now use the actual multiplication factor for 2+ stacks, so I got rid of the language-specific if/else and I think the ZHS strings won't even need to change.

src/main/resources/anniv5Resources/localization/eng/serpentinepack/Powerstrings.json
```diff
@@ -14,8 +14,8 @@
     "NAME": "Tainted",
     "DESCRIPTIONS": [
       "Doubles any debuffs you apply to this enemy.",
-      "Doubles any debuffs you apply to this enemy #b",
```
modargo, line 17: The old description wasn't even accurate -- it multiplied by 2x, but the way to double something _x_ times is to multiply by 2<sup>x</sup>. This is probably why the ZHS string was set up differently, to give a more accurate description.

PR #467, opened 2023-03-02
Bulwark pack: add a new defensively-focused pack

modargo:


src/main/resources/anniv5Resources/localization/eng/UIstrings.json
```diff
@@ -83,6 +83,19 @@
...
+  "${ModID}:BulwarkPack": {
+    "TEXT": [
+      "Bulwark",
+      "The greatest defensive tools of the Spire, together for the first time in one place.",
+      "modargo"
+    ]
+  },
+  "${ModID}:BulwarkPackHat": {
+    "TEXT": [
+      "TBD",
```
modargo, line 95: Not sure how we've filled in hats for new packs like this, this is the one thing missing from the pack.

src/main/resources/anniv5Resources/summaries/tags.txt
```diff
@@ -11,6 +11,7 @@ BellordPack	None
 BitingColdPack	Debuffs
 BlacksmithsPack	None
 BoardGamePack	None
+BulwarkPack	Exhaust
```
modargo, line 14: Feel No Pain is strong enough support that I put the Exhaust tag on the pack. Could remove it if we think it's misleading, but even if the other 9 cards didn't exist, adding FNP to the card pool would still be one of the strongest things you can do when playing with packs like Distortion or Psychic.

src/main/resources/anniv5Resources/summaries/ratings.txt
```diff
@@ -11,6 +11,7 @@ BellordPack	2	3	2	4	1
 BitingColdPack	3	2	2	3	2
 BlacksmithsPack	3	3	4	4	4
 BoardGamePack	4	3	2	4	2
+BulwarkPack	2	5	1	4	4
```
modargo, line 14: The first 5 in Defense! Also a 1 in Support, because it doesn't even have anything that draws cards or makes energy, which feels appropriate for a very focused pack like this one.

PR #468, opened 2023-03-03
- added intrigue pack

AlexMdle:


src/main/java/thePackmaster/cards/intriguepack/Exultation.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        this.applyPowers();
+        Wiz.atb(new AbstractGameAction() {
+            @Override
+            public void update() {
+                for(AbstractCard c : Wiz.p().hand.group)
+                {
+                    if (promote(c) && c != Exultation.this)
```
erasels, line 27: Wouldn't this promote itself but just not give the block? Not what the effect says.

src/main/java/thePackmaster/cards/intriguepack/Exultation.java
```diff
@@ -0,0 +1,53 @@
...
+                        Wiz.doBlk(magicNumber);
```
erasels, line 28: Most effects based block gains collect the amount of block and then give it to you once so they don't scale out of control with Dex. Might want to do the same here.

src/main/java/thePackmaster/cards/intriguepack/Exultation.java
```diff
@@ -0,0 +1,53 @@
...
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class Exultation extends AbstractIntrigueCard {
+    public final static String ID = makeID("Exultation");
+
+    public Exultation() {
+        super(ID, 2, CardType.SKILL, CardRarity.UNCOMMON, CardTarget.SELF);
+        magicNumber = baseMagicNumber = 3;
+        baseBlock = 0;
```
erasels, line 17: Hmm, this is a weird edgecase. If you do this you'll mess with some stuff that expects block cards to use their block instead of overwriting it. Might be better to leave baseBlock at -1 until needed since you overwrite it anyway.

src/main/java/thePackmaster/cards/intriguepack/Shakedown.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public Shakedown() {
+        super(ID, 1, CardType.ATTACK, CardRarity.COMMON, CardTarget.ENEMY);
+        baseDamage = DAMAGE;
+    }
+
+    public boolean hasRarity(CardRarity rar) {
+        Iterator var1 = Wiz.p().hand.group.iterator();
+
+        while(var1.hasNext()) {
```
erasels, line 30: Why do you do this? Use a for each loop like a normal person you monster

src/main/java/thePackmaster/cards/intriguepack/Shakedown.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        return false;
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+                Wiz.atb(new DamageAction(m, new DamageInfo(p, damage, damageTypeForTurn), AbstractGameAction.AttackEffect.BLUNT_LIGHT));
+
+                if (hasRarity(CardRarity.COMMON) || hasRarity(CardRarity.BASIC) || hasRarity(CardRarity.SPECIAL))
```
erasels, line 44: Yeet this into an action. Also would be way more efficient to just iterate through the list once and set three booleans (hasMundane, hasRare, etc)

src/main/java/thePackmaster/cards/intriguepack/Spotlight.java
```diff
@@ -0,0 +1,23 @@
...
+import thePackmaster.actions.intriguepack.SpotlightAction;
+import thePackmaster.util.Wiz;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class Spotlight extends AbstractIntrigueCard {
+    public final static String ID = makeID("Spotlight");
+
+    public Spotlight() {
+        super(ID, 0, CardType.SKILL, CardRarity.UNCOMMON, CardTarget.NONE);
```
erasels, line 14: Seems extremely powerful at 0 cost and non exhaust when compared to Hologram or Seek. Might want to increase the cost to 1

src/main/java/thePackmaster/cards/intriguepack/Topple.java
```diff
@@ -0,0 +1,66 @@
...
+        super.applyPowers();
+        this.baseDamage = realBaseDamage;
+        this.isDamageModified = this.damage != this.baseDamage;
+    }
+
+    public int countCards() {
+        int counter = 0;
+        Iterator var1 = AbstractDungeon.actionManager.cardsPlayedThisCombat.iterator();
+
+        while(var1.hasNext()) {
```
erasels, line 47: nopers

src/main/java/thePackmaster/patches/intriguepack/CopyRarity.java
```diff
@@ -0,0 +1,19 @@
...
+
+@SpirePatch(
+        clz = AbstractCard.class,
+        method = "makeStatEquivalentCopy"
+)
+public class CopyRarity {
+    @SpirePostfixPatch
+    public static AbstractCard transferIt(AbstractCard __result, AbstractCard __instance)
+    {
+        __result.rarity = __instance.rarity;
```
erasels, line 16: This could be problematic with BasicWave from Basics pack, have you tested how they interact?

erasels (changes requested):

PR #469, opened 2023-03-05
Nerfs for Basic pack

GameboyJ:
Nerfs for basic pack.

Conjure Defends now generates a Defend(+) that is played X times, or in other words, X-1 additional times.
Simplify now upgrades to upgrade twice instead of drawing 2
Lose It now increase the cards drawn by rummages by 1 instead of reducing their cost to 1.
decided to leave Ockhams Razor untouched, as the other nerfs are an indirect nerf.

also made some GrandOpening cards glow when they should.


src/main/java/thePackmaster/cards/grandopeningpack/PlannedDefense.java
```diff
@@ -37,6 +39,21 @@ public void use(AbstractPlayer p, AbstractMonster m) {
         }
     }
 
+    @Override
+    public void triggerOnGlowCheck() {
+        boolean innatePlayed = false;
+        this.glowColor = AbstractCard.BLUE_BORDER_GLOW_COLOR.cpy();
```
erasels, line 45: Please don't do this, you're creating a new color every time even when it might be overriden by gold. Instead just add an else to the if at the end and then set the glowColor to blue there

src/main/java/thePackmaster/cards/grandopeningpack/SecondStrike.java
```diff
@@ -29,6 +31,14 @@ public void use(AbstractPlayer abstractPlayer, AbstractMonster abstractMonster)
         addToBot(new DamageAction(abstractMonster, new DamageInfo(abstractPlayer, this.damage, DamageInfo.DamageType.NORMAL), AbstractGameAction.AttackEffect.SLASH_DIAGONAL));
     }
 
+    @Override
+    public void triggerOnGlowCheck() {
+        this.glowColor = AbstractCard.BLUE_BORDER_GLOW_COLOR.cpy();
+        if (GameActionManager.turn > 1) {
+            this.glowColor = AbstractCard.GOLD_BORDER_GLOW_COLOR.cpy();
+        }
```
erasels, line 39: Please just add an else here to set it to blue

src/main/java/thePackmaster/cardmodifiers/basicspack/MagicModifier.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public MagicModifier(int value, boolean secondMagic){
+        this.value = value;
+        this.secondMagic = secondMagic;
+    }
+
+    public boolean shouldApply(AbstractCard card) {
+        if(secondMagic && card instanceof AbstractPackmasterCard){
+            return ((AbstractPackmasterCard) card).secondMagic != 0;
+        } else return card.magicNumber != 0;
```
erasels, line 26: Should this be != -1? 0 means it was changed, -1 means its probably not being used.

src/main/java/thePackmaster/cards/grandopeningpack/Improvise.java
```diff
@@ -37,6 +37,21 @@ public void use(AbstractPlayer abstractPlayer, AbstractMonster abstractMonster)
...
+        boolean holdingInnate = false;
+        this.glowColor = AbstractCard.BLUE_BORDER_GLOW_COLOR.cpy();
+        for (AbstractCard c : AbstractDungeon.player.hand.group) {
+            if (c.isInnate && c != this) {
+                holdingInnate = true;
+                break;
+            }
+        }
+        if (holdingInnate) {
+            this.glowColor = AbstractCard.GOLD_BORDER_GLOW_COLOR.cpy();
```
erasels, line 51: Same as the others

src/main/resources/anniv5Resources/localization/eng/Cardstrings.json
```diff
@@ -9,7 +9,7 @@
   },
   "${ModID}:Rummage": {
     "NAME": "Rummage",
-    "DESCRIPTION": "{@@}Deal !D! damage. NL Draw a card and reduce its cost by !M! this turn."
+    "DESCRIPTION": "{@@}Deal !D! damage. NL Draw !${ModID}:m2! {@@}{!${ModID}:m2!|@=card|>1=cards} and reduce {@@}{!${ModID}:m2!|@=its|>1=their} cost by !M! this turn."
```
erasels, line 12: Don't forget to overwrite the zhs strings so mwalls can easily see what to change

erasels (changes requested):

src/main/resources/anniv5Resources/localization/eng/Cardstrings.json
GameboyJ: what do you mean with overwite the zhs string?

src/main/java/thePackmaster/cards/grandopeningpack/PlannedDefense.java
GameboyJ: I copied this from Ripped Pecs, which also first makes it glow blue and then gold, i can change that too if you want, or is there some difference i am not seeing?

PR #480, opened 2023-03-10
Jockey Pack

DarkVexon:
Ride a Horse to victory with the Jockey pack, which specializes in playing and drawing as many cards as possible in a single turn!

Also tweaked a few of my other cards.
- Midnight Strike 12-card damage to 36(48). Too easy to trivialize a fight.
- Like Clockwork cost down to 0, Block down to 4(6). Too strong when played, not fun enough to play.

Fixed a bug that would cause Enough Talk and Conglaturation to trigger if they were possible in the previous combat's final turn.

Added visual effects and sound effects when you trigger Ripped Pecs' Synergy. (same as angery champ)


erasels:
Gonna save this for the next release.

PR #487, opened 2023-03-12
- goddess of explosions tweaks

AlexMdle:


erasels:
Oh, please switch the ZHS strings around as well

AlexMdle:
It doesn't have zhs strings.

PR #489, opened 2023-03-13
Pyromaniac streamline

Laugic:


Laugic:
Also fixed Dice removal timing so it works with sword boomerang now

PR #501, opened 2023-03-14
Compendium: organize the Packmaster's compendium tab to cleanly show each pack in two rows of five cards

modargo:
To understand what's going on with the code I wrote you'll need to look at https://github.com/daviscook477/BaseMod/blob/1a8a344ecfde07613d29f39ff52b380cf50196f9/mod/src/main/java/basemod/patches/com/megacrit/cardcrawl/screens/compendium/CardLibraryScreen/EverythingFix.java and https://github.com/daviscook477/BaseMod/blob/1a8a344ecfde07613d29f39ff52b380cf50196f9/mod/src/main/java/basemod/patches/com/megacrit/cardcrawl/screens/mainMenu/ColorTabBar/ColorTabBarFix.java. The compendium tab for modded characters only exists in BaseMod, so most of what I've written is patches on top of BaseMod's patches combined with a bit of digging into the internals of CardLibraryScreen.

### DIFF
--- a/src/main/java/thePackmaster/patches/CompendiumPatches.java
+++ b/src/main/java/thePackmaster/patches/CompendiumPatches.java
@@ -1,10 +1,13 @@
 package thePackmaster.patches;
 
 import basemod.ReflectionHacks;
+import basemod.patches.com.megacrit.cardcrawl.screens.compendium.CardLibraryScreen.EverythingFix;
 import com.evacipated.cardcrawl.modthespire.lib.*;
 import com.megacrit.cardcrawl.cards.AbstractCard;
+import com.megacrit.cardcrawl.cards.CardGroup;
 import com.megacrit.cardcrawl.core.CardCrawlGame;
 import com.megacrit.cardcrawl.core.Settings;
+import com.megacrit.cardcrawl.helpers.CardLibrary;
 import com.megacrit.cardcrawl.helpers.Hitbox;
 import com.megacrit.cardcrawl.localization.UIStrings;
 import com.megacrit.cardcrawl.screens.compendium.CardLibSortHeader;
@@ -14,9 +17,12 @@ import javassist.CtBehavior;
 import javassist.expr.ExprEditor;
 import javassist.expr.NewExpr;
 import thePackmaster.SpireAnniversary5Mod;
+import thePackmaster.ThePackmaster;
 
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.List;
+import java.util.stream.Collectors;
 
 public class CompendiumPatches {
     public static UIStrings uiStrings = CardCrawlGame.languagePack.getUIString(SpireAnniversary5Mod.makeID("Compendium"));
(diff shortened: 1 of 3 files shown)


src/main/java/thePackmaster/patches/CompendiumPatches.java
```diff
@@ -131,4 +138,28 @@ public static void packSort(CardLibSortHeader input, boolean isAscending) {
...
     }
+
+    @SpirePatch2(clz = EverythingFix.Initialize.class, method = "Insert")
+    public static class ShowBaseGameCards {
+        @SpirePostfixPatch
+        public static void showBaseGameCards(Object __obj_instance) {
+            if (EverythingFix.Fields.cardGroupMap.containsKey(ThePackmaster.Enums.PACKMASTER_RAINBOW)) {
+                CardGroup group = EverythingFix.Fields.cardGroupMap.get(ThePackmaster.Enums.PACKMASTER_RAINBOW);
+                List<AbstractCard> baseGameCards = CardLibrary.getAllCards().stream()
+                        .filter(c -> (c.color == AbstractCard.CardColor.RED || c.color == AbstractCard.CardColor.GREEN || c.color == AbstractCard.CardColor.BLUE || c.color == AbstractCard.CardColor.PURPLE))
```
modargo, line 149: In theory we could allow people to add colorless cards to packs as well, but to my knowledge nobody has done that and I thought it was better to start with a more restrictive check here.

src/main/java/thePackmaster/patches/RenderBaseGameCardPackTopTextPatches.java
```diff
@@ -1,26 +1,31 @@
...
+import com.megacrit.cardcrawl.screens.compendium.CardLibraryScreen;
+import com.megacrit.cardcrawl.screens.mainMenu.ColorTabBar;
+import com.megacrit.cardcrawl.screens.mainMenu.MainMenuScreen;
 import thePackmaster.ThePackmaster;
 import thePackmaster.cards.AbstractPackmasterCard;
 import thePackmaster.packs.AbstractCardPack;
 import thePackmaster.util.Wiz;
 
 @SpirePatch2(clz = AbstractCard.class, method = "render", paramtypez = {SpriteBatch.class})
+@SpirePatch2(clz = AbstractCard.class, method = "renderInLibrary", paramtypez = {SpriteBatch.class})
```
modargo, line 24: As usual, any code dealing with card rendering has to cover the three cases of normal rendering, rendering in the compendium (what I added here), and rendering in single card view (the other patch file I modified).

src/main/java/thePackmaster/patches/RenderBaseGameCardPackTopTextPatches.java
```diff
@@ -49,4 +54,25 @@ public static void patch(AbstractCard __instance, SpriteBatch sb) {
...
+        return !Settings.hideCards
+                && c.getClass().getSuperclass().equals(AbstractCard.class)
+                && (isInPackmasterRun() || isInPackmasterCardLibraryScreen());
+    }
+
+    private static boolean isInPackmasterRun() {
+        return AbstractDungeon.player != null && AbstractDungeon.player.chosenClass == ThePackmaster.Enums.THE_PACKMASTER;
+    }
+
+    private static boolean isInPackmasterCardLibraryScreen() {
```
modargo, line 68: This function is the new code -- the other two functions above are just me cleaning up and reorganizing the previously existing logic so that it was easier to follow.

The logic here isn't too bad, but it all piggy-backs on the BaseMod ColorTabBarFix patches, so to understand it you really need to understand what BaseMod is doing.

src/main/java/thePackmaster/patches/SingleCardViewBorderTextPatch.java
```diff
@@ -39,7 +39,7 @@ public static void Postfix(SingleCardViewPopup __instance, SpriteBatch sb) {
             AbstractCard card = (AbstractCard)cardField.get(__instance);
             if (card instanceof AbstractPackmasterCard) {
                 ((AbstractPackmasterCard)card).renderBorderText(sb, Settings.WIDTH / 2.0F, (float)yField.get(__instance), yOffsetBase, drawScale);
-            } else if (AbstractDungeon.player != null && AbstractDungeon.player.chosenClass == ThePackmaster.Enums.THE_PACKMASTER && card.getClass().getSuperclass().equals(AbstractCard.class)) {
+            } else if (RenderBaseGameCardPackTopTextPatches.shouldShowPackName(card)) {
```
modargo, line 42: Previously this had the same condition check as the non-SCV version. That was fine when it wasn't too complex, but with the additional complexity that I introduced, I centralized the logic.

modargo:
Leaving this up for you to merge since it's a bit tricky and it's worth you knowing this logic, in case it needs to be fixed or changed in the future (e.g. after Vex's colorless pack).

PR #515, opened 2023-03-17
Compendium: hide and disable the sort by pack button on tabs other than the Packmaster

modargo:
In between working on other stuff, I realized that the work I'd done for the cleaned up compendium would also let me fix another thing that's always bothered me: the sort by packs button showing up for other tabs in the compendium. I think nobody really wanted to implement the logic "are we in the Packmaster's compendium tab" (which is more work than it might seem to figure out) just to fix just to fix a minor UI issue.

However, I already had to write that logic for that as part of cleaning up the compendium. With the function `isInPackmasterCardLibraryScreen` available to use, hiding the sort by packs button from other tabs only needed tweaks to a few existing patches and a few new patches that were relatively straightforward.

To summarize the changes:
* We use two new patches to prevent rendering or clicking the sort by packs button when you're on other tabs.
* When you switch to a non-Packmaster tab with sort by packs active, we ignore it, meaning that the tab acts exactly like it does when you first enter it with no sort selected.
* When you switch to a non-Packmaster tab with sort by packs active, then switch back to the Packmaster without changing sorting, we still sort by packs.
* The UI for other tabs will still look slightly different than normal, since the code that shrinks size of each sort button in order to fit in the new one is still active (even if the new button isn't there for other tabs). This is for the best, since it would be annoying and odd if the buttons resized themselves when you switched tabs.

### DIFF
--- a/src/main/java/thePackmaster/patches/CompendiumPatches.java
+++ b/src/main/java/thePackmaster/patches/CompendiumPatches.java
@@ -2,9 +2,11 @@ package thePackmaster.patches;
 
 import basemod.ReflectionHacks;
 import basemod.patches.com.megacrit.cardcrawl.screens.compendium.CardLibraryScreen.EverythingFix;
+import com.badlogic.gdx.graphics.g2d.SpriteBatch;
 import com.evacipated.cardcrawl.modthespire.lib.*;
 import com.megacrit.cardcrawl.cards.AbstractCard;
 import com.megacrit.cardcrawl.cards.CardGroup;
+import com.megacrit.cardcrawl.cards.DescriptionLine;
 import com.megacrit.cardcrawl.core.CardCrawlGame;
 import com.megacrit.cardcrawl.core.Settings;
 import com.megacrit.cardcrawl.helpers.CardLibrary;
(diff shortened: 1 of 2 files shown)


src/main/java/thePackmaster/patches/CompendiumPatches.java
```diff
@@ -67,7 +72,7 @@ public int[] Locate(CtBehavior ctMethodToPatch) throws Exception {
     public static class CustomOrdering {
         @SpirePostfixPatch
         public static void catchSort(CardLibSortHeader __instance, SortHeaderButton button, boolean isAscending) {
-            if (button == packButton) {
+            if (button == packButton && RenderBaseGameCardPackTopTextPatches.isInPackmasterCardLibraryScreen()) {
```
modargo, line 75: I believe that since you can't click the sort by packs button on other tabs, this check may be unnecessary (since we can assume that when you hit this code, you're on the Packmaster tab). I still thought it was prudent to have the check here too, as an extra safeguard in case we missed some way of hitting this, and to make it clear that the intent of this is to only work for the Packmaster tab.

src/main/java/thePackmaster/patches/RenderBaseGameCardPackTopTextPatches.java
```diff
@@ -65,7 +65,7 @@ private static boolean isInPackmasterRun() {
         return AbstractDungeon.player != null && AbstractDungeon.player.chosenClass == ThePackmaster.Enums.THE_PACKMASTER;
     }
 
-    private static boolean isInPackmasterCardLibraryScreen() {
+    public static boolean isInPackmasterCardLibraryScreen() {
```
modargo, line 68: This set of functions could be moved to some common location, now that they're being used by a few different patches.

I decided not to do that because I don't see these functions as more broadly useful. They help with a small and specific set of functionality, and I don't think there will be much more code written that would want to use this.

PR #524, opened 2023-03-19
Character event: switch the curse given by the Learn Dark Arts option to a new Packmaster-specific curse. Hi Vex!

modargo:
This PR doesn't have art yet, but I'm putting it up so that we have flexibility to do what we want with this change, regardless of whether I'm around and able to be actively doing stuff with it. From my perspective, the main options are:
* Merge this in and release even without art (the card art roller gives it fairly appropriate art, see below)
* Merge this in and add art (if Vex has it ready)
* Wait until after this release (if the art isn't ready, we don't want to release this change without art, and we don't want to delay the release for this)

I have very little preference between the options, beyond a general like for getting things merged in.

image

### DIFF
--- a/src/main/java/thePackmaster/cards/Vexed.java
+++ b/src/main/java/thePackmaster/cards/Vexed.java
@@ -1,0 +1,37 @@
+package thePackmaster.cards;
+
+import com.evacipated.cardcrawl.mod.stslib.fields.cards.AbstractCard.SoulboundField;
+import com.megacrit.cardcrawl.actions.common.ExhaustSpecificCardAction;
+import com.megacrit.cardcrawl.cards.AbstractCard;
+import com.megacrit.cardcrawl.characters.AbstractPlayer;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.monsters.AbstractMonster;
+import thePackmaster.patches.odditiespack.PackmasterFoilPatches;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class Vexed extends AbstractPackmasterCard {
+    public final static String ID = makeID("Vexed");
+    private static final int COST = -2;
+
+    public Vexed() {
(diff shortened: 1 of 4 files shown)


src/main/resources/anniv5Resources/localization/eng/Cardstrings.json
```diff
@@ -19,5 +19,9 @@
     "NAME": "Conjure Pack",
     "DESCRIPTION": "Choose 1 of !M! Packs available to your run. NL At the start of the next 5 turns, add 2 of its cards to your hand.",
     "UPGRADE_DESCRIPTION": "Innate. NL Choose 1 of !M! Packs available to your run. NL At the start of the next 5 turns, add 2 of its cards to your hand."
+  },
+  "${ModID}:Vexed": {
+    "NAME": "Vexed",
+    "DESCRIPTION": "Unplayable. NL Cannot be removed from your deck. NL While in hand, Exhaust this when you play a *Foil card."
```
modargo, line 25: I took the "While in hand" part of the wording from Pain, which is one of the only base game cards with an effect like this (Normality also uses "While in hand" though that effect is constant rather than triggered).

src/main/java/thePackmaster/cards/Vexed.java
```diff
@@ -0,0 +1,37 @@
...
+import com.megacrit.cardcrawl.actions.common.ExhaustSpecificCardAction;
+import com.megacrit.cardcrawl.cards.AbstractCard;
+import com.megacrit.cardcrawl.characters.AbstractPlayer;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.monsters.AbstractMonster;
+import thePackmaster.patches.odditiespack.PackmasterFoilPatches;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class Vexed extends AbstractPackmasterCard {
```
modargo, line 13: Note that this being unplayable is taken care of by the default logic for curses in `AbstractCard.canUse`.

PR #529, opened 2023-03-20
General: the Merchant now has a few things to say to the Packmaster

modargo:
Here's the next little easter egg from me -- a few custom Merchant strings for the Packmater. These will occasionally show up in the things that Merchant says while you're in a shop. Most people won't notice, but I think it will be cool for people who do.

For context, I've included the normal things the Merchant says below (the first set are pre-act 4, the second set is act 4). The pre-act 4 messages are small talk or statements about the merchant or the shop. The act 4 messages are a bit stranger, ranging from light-hearted to ominous ("I used to be like you." stands out for that).

The Packmaster-specific messages I've written are intended to follow the pattern set by the base game. For pre-act 4, a few references to things about the Packmaster and small talk. For act 4, one slightly strange question and one strange and ominous statement.
```
"TEXT": [
  "Hi! NL Buy something! NL Buy buy buy!",
  "I like your haircut",
  "Do you have the time?",
  "Are you a dog person or a cat person?",
  "You look ~dangerous,~ NL hehe...",
  "Take your time... NL ..... NL ... or not",
  "Do you like this rug? NL It's not for sale",
  "Support small business!",
  "Masks are cool, I would know!",
  "Hmm... NL ...hmm... NL Oh!",
  "Stay awhile and listen!",
  "It's dangerous to go alone! Give me all your gold!",
  "Buy somethin'",
  "~*YAWN*~",
  "I like gold.",
  "My favorite color is blue. NL How bout you?"
],
"OPTIONS": [
  "How did I get here? NL Beats me.",
  "Last chance to buy somethin'! NL ~*wink*~ ~*wink*~",
  "..... NL .... NL .........",
  "I'm a cat person.",
  "No rush. NL No rush.",
  "Now isn't this @exciting!?@",
  "I used to be like you.",
  "Heading up?"
]
```

### DIFF
--- a/src/main/java/thePackmaster/patches/MerchantSpeechPatch.java
+++ b/src/main/java/thePackmaster/patches/MerchantSpeechPatch.java
@@ -1,0 +1,43 @@
+package thePackmaster.patches;
+
+import basemod.ReflectionHacks;
+import com.evacipated.cardcrawl.modthespire.lib.SpirePatch;
+import com.evacipated.cardcrawl.modthespire.lib.SpirePatch2;
+import com.evacipated.cardcrawl.modthespire.lib.SpirePostfixPatch;
+import com.megacrit.cardcrawl.core.CardCrawlGame;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.dungeons.TheEnding;
+import com.megacrit.cardcrawl.shop.Merchant;
+import com.megacrit.cardcrawl.shop.ShopScreen;
+import thePackmaster.SpireAnniversary5Mod;
+import thePackmaster.ThePackmaster;
+
+import java.util.ArrayList;
+import java.util.Collections;
+
+public class MerchantSpeechPatch {
+    private static final String[] speech = CardCrawlGame.languagePack.getUIString(SpireAnniversary5Mod.makeID("MerchantSpeech")).TEXT;
+    private static final String[] endingSpeech = CardCrawlGame.languagePack.getUIString(SpireAnniversary5Mod.makeID("MerchantEndingSpeech")).TEXT;
+
+    @SpirePatch2(clz = Merchant.class, method = SpirePatch.CONSTRUCTOR, paramtypez = { float.class, float.class, int.class })
+    public static class MerchantPatch {
+        @SpirePostfixPatch
+        public static void addEntries(Merchant __instance) {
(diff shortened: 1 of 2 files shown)


src/main/java/thePackmaster/patches/MerchantSpeechPatch.java
```diff
@@ -0,0 +1,43 @@
...
+    private static final String[] speech = CardCrawlGame.languagePack.getUIString(SpireAnniversary5Mod.makeID("MerchantSpeech")).TEXT;
+    private static final String[] endingSpeech = CardCrawlGame.languagePack.getUIString(SpireAnniversary5Mod.makeID("MerchantEndingSpeech")).TEXT;
+
+    @SpirePatch2(clz = Merchant.class, method = SpirePatch.CONSTRUCTOR, paramtypez = { float.class, float.class, int.class })
+    public static class MerchantPatch {
+        @SpirePostfixPatch
+        public static void addEntries(Merchant __instance) {
+            if (AbstractDungeon.player != null && AbstractDungeon.player.chosenClass == ThePackmaster.Enums.THE_PACKMASTER) {
+                ArrayList<String> idleMessages = ReflectionHacks.getPrivate(__instance, Merchant.class, "idleMessages");
+                Collections.addAll(idleMessages, AbstractDungeon.id.equals(TheEnding.ID) ? endingSpeech : speech);
```
modargo, line 28: One funny thing that I noticed: the Merchant uses the pre-act 4 messages for alternate act 4s. ActLikeIt doesn't patch that code or anything. To maintain consistency I've written the same check here, since otherwise you'd get a weird mix of messages.

src/main/java/thePackmaster/patches/MerchantSpeechPatch.java
```diff
@@ -0,0 +1,43 @@
...
+            }
+        }
+    }
+
+    @SpirePatch2(clz = ShopScreen.class, method = "init", paramtypez = { ArrayList.class, ArrayList.class })
```
modargo, line 33: These two patches are almost the same, because the base game sets up these messages in two places that are almost the same. Even though there are ways the patches could be combined, they all end up harder to understand than this. (The patch body has to be slightly different due to the ReflectionHacks call -- without that, it might be possible to combine these by just having it take `Object __instance`.)

modargo:
Leaving this one up for you to merge since its a new feature.

PR #530, opened 2023-03-21
Instant Death pack

Alchyr:
Done
Needs chinese localization

image
image

Precision - Vigor, but the damage increase lasts the rest of combat on the card that's played
Nail gives less precision than in images
Upgraded needle has +1 damage

Current ratings:
4 Offense
3 Defense
3 Support
2 Frontload
3 Scaling

4 Offense might be more accurate, but having less than 5 feels wrong


erasels:
We don#t have russion localization, do you mean chinese?

modargo:
Dropping by since you were interested in my thoughts on ratings. I'd probably put offense as 4 and frontload as 2. I'll write out some thoughts on why.

For offense, the pack is offensively focused but not exclusively so -- I'd label only 5 of the 10 cards as offense cards (the four attacks and Path to Victory since it gives Precision). Curtain and Lock are a mix of support and defense, Cloud is defense, Another Moment is defense with exhaust synergy, and Mirror is support (that's how copy/duplicate effects are classified). 

For frontload, I think a below average rating is right, but the pack isn't entirely lacking frontload or anything. In fact I think the commons and uncommons will do an okay job.
* Curtain is essentially Leap -- 1 energy for 9 block (and if you don't need to block, you can keep around some of the cards you see). Only issue it will have is when your deck is empty or you scry a card you really want to keep.
* Needle is more conditional but on average you should be able to get 3 Precision from it, so we can look at it as typically being 9 damage (slightly delayed, but also more damage if you play the attack that gets the Precision buff more than once). Maybe a bit lower than you'll get from strong common attacks, but decent.
* Nail is Slice with upside if you Exhaust it. Again, decent.
* Thread is Sweeping Beam++ (1 more damage, the draw is always a skill, and it shuffles itself back in).
* Cloud also contributes to frontload in the endgame since the benefit from it scales with how much damage enemies do. In act 3 and 4, the 2 energy for Cloud will easily be equivalent to 20+ block for elites and bosses. Can't give it full credit for frontload since its overcosted for the early game, but I try to look at this stuff with a full-game perspective.
* The other commons and uncommons either contribute less frontload (Lock) or are harder to evaluate (Mirror).

3s in Defense and Support seem reasonable. A 3 for scaling is also reasonable, though the existence of Cloud, Another Moment, and Slash Draw gives me some pause. It makes me wonder if the scaling those enable will be stronger than anticipated, both through the straightforward use of the cards and by scaling-through-exhaust. Also Mirror to some extent, through the potential of scrying to get one of your best cards on top (like a key scaling card) and then playing Mirror+ to get an extra copy (I say Mirror+ because you still need the energy to play the card you get, and I feel that will be hard to set up with unupgraded Mirror).

FWIW, just looking at the cards, minor feedback:
* I'd be inclined to give Needle +1 damage on upgrade (compare to Cut Through Fate, which gets +2 damage and +1 Scry)
* Something like 6 (8) Precision might be more balanced for Nail, with the added benefit of making it symmetrical with the damage.
* I worry a bit about Cloud since multiplicative damage reduction is an inherently powerful scaling effect. This is the exact type of effect where my gut reaction is to say that maybe it should exhaust. To be clear, I'm fine with giving the current version a try, just saying that we should keep an eye on it. Someone is totally going to get Snecko Eye and a few copies of Cloud and have it up every turn.
* The rares all offer interesting puzzles to take advantage of their scaling. Makes it a bit tricky to judge their strength so I'll be curious to see how they play.

having a bunch of cards left in your hand and putting them all on top of your deck is either sacrificing plays you could have made, or it means the fight is already

src/main/java/thePackmaster/cards/instadeathpack/PathToVictory.java
```diff
@@ -0,0 +1,39 @@
...
+    public PathToVictory() {
+        super(ID, 1, CardType.SKILL, CardRarity.RARE, CardTarget.NONE);
+
+        this.baseMagicNumber = this.magicNumber = 3;
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        int vigor = magicNumber;
+        addToBot(new HandSelectAction(69420, (c)->true, (cards)->{
```
erasels, line 26: nice

src/main/java/thePackmaster/powers/instadeathpack/AnotherMomentPower.java
```diff
@@ -0,0 +1,57 @@
...
+    {
+        super(POWER_ID, NAME, PowerType.BUFF, false, owner, amount);
+    }
+
+    public void atStartOfTurn() {
+        this.exhaustedThisTurn = 0;
+    }
+
+    public void onUseCard(AbstractCard card, UseCardAction action) {
+        if (card.type == AbstractCard.CardType.SKILL) {
```
erasels, line 31: Might've been good as a two amount power?

src/main/java/thePackmaster/powers/instadeathpack/LockPower.java
```diff
@@ -0,0 +1,68 @@
...
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.core.CardCrawlGame;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.helpers.FontHelper;
+import com.megacrit.cardcrawl.powers.AbstractPower;
+import thePackmaster.SpireAnniversary5Mod;
+import thePackmaster.powers.AbstractPackmasterPower;
+
+
+public class LockPower extends AbstractPackmasterPower implements NonStackablePower {
```
erasels, line 17: Would be nice if this could use Nelly's new icon system (like KillQueen) to show the card it's on.

erasels (changes requested):

src/main/java/thePackmaster/powers/instadeathpack/LockPower.java
Alchyr: 👍

src/main/java/thePackmaster/powers/instadeathpack/AnotherMomentPower.java
Alchyr: eh maybe but I think it's fine as is, functions the same as echo form

src/main/java/thePackmaster/cards/instadeathpack/PathToVictory.java
Alchyr: 💯

PR #531, opened 2023-03-22
Beep Boop

bkerns724:


erasels:
I also think Ambush was more interesting than Specialist, sad to see you make this decision.

bkerns724:
Ambush is staying, Training isn't.  I just haven't had a run where I enjoyed how training works.  Also even with just ambush the pack seems to lean towards evoke more than passive, which is fine if not taken too far, but training + ambush together pushes that too far.

bkerns724:
Okay, I think I trimmed out all the unused code from my pack.

PR #533, opened 2023-03-22
support for custom number of pack options

Alchyr:
allows choice of any number instead of just 1 of 3

requires addition to config for users to actually set it
zhs localization update
might need to modify text more to make it flexible for zhs?

I recommend just having the option between 2-5
or maybe just 3 Or 5

More than that doesn't really fit

to change value requires
change `SpireAnniversary5Mod.PACKS_PER_CHOICE`
call `MainMenuUIPatch.updateChoiceCount`

Note - also prevents softlock if most packs are disabled


erasels:
Am I to understand you don't want to add the necessary UI component to change between the 2? Just to be sure I'm understanding you correctly

Alchyr:
yes

I have made a mod config 0 times in total :)

PR #537, opened 2023-03-23
New Pack: Colorless Creations!

DarkVexon:
Adds a new pack centered around using and creating Colorless cards. All sorts of wacky new effects, awesome Vex art and more!

Also:
-Fixes a bug where Handy Haversack and Bag of Holding were using addToTop, causing bad interactions with other modded content.
-ALL Packmaster tokens are now Colorless and display as such.
-Adds non-duplicate art for the Emptiness card.

WOOOO IT'S DONE


src/main/resources/anniv5Resources/summaries/ratings.txt
```diff
@@ -13,6 +13,7 @@ BlacksmithsPack	3	3	4	4	4
 BoardGamePack	4	3	2	4	2
 BulwarkPack	2	5	1	4	4
 ClawPack	5	2	1	2	5
+ColorlessPack	4	3	3	2	5
```
modargo, line 16: I'm a bit doubtful of the 5 in scaling here. The pack looks to have decent scaling, but it's not as straightforward as other packs with a 5 in scaling like Claw, Distortion, or Poison. For what it's worth, having played with your Secret Level pack, that 5 in scaling is probably also too high. That pack provides reasonably good scaling but it doesn't outclass the various other packs that have good scaling and are rated as a 4.

This might come from a bit of overestimating the value of the scaling effects you've created. A couple sources of Strength or Dexterity gain alongside a few other ways to scale (e.g. Prism for this pack, The Donut for Secret Level) is good and will make a pack above average, but it doesn't reach the same consistency and speed as most of the packs rated 5 in scaling.

Let's change this to a 4 in scaling and also change Secret Level to have a 4 in scaling.

modargo (changes requested):
Looks pretty good overall. A few bugs to fix and other small changes to make.

modargo:
I started a run with this pack and just had the fun experience of getting The Prism from an attack potion. Talk about value!

src/main/java/thePackmaster/cards/colorlesspack/ThePrism.java
```diff
@@ -0,0 +1,79 @@
...
+    public final static String ID = makeID("ThePrism");
+    private boolean showPreview;
+    // intellij stuff attack, all_enemy, rare, 8, 1, , , 8, 1
+
+    public ThePrism(boolean showPreview) {
+        super(ID, 3, CardType.ATTACK, CardRarity.RARE, CardTarget.ALL_ENEMY);
+        this.showPreview = showPreview;
+        baseDamage = 8;
+        baseMagicNumber = magicNumber = 8;
+        selfRetain = true;
```
modargo, line 27: It looks like this is still marked as retaining (from an earlier draft of the card where it added to your hand, I assume). Should remove this.

modargo:
Gameplay feedback for Ghost: the optimal way to play with Ghost is to look through your hand and draw pile at the start of each combat to work out which card(s) you have an extra of. Even if you don't have a perfect memory, you just need to compare to your deck.

In other words, Ghost is a card where the optimal play is tedious, which is generally bad by principles of game design and is something that Slay the Spire has specifically tried to avoid (thus why healing cards can't be generated during combat). Regardless of whether this was the "intended" way to play with Ghost, it's the way good players are going to approach it when they realize what the card does. That makes it something we should think about for designing these cards, and in my view it makes the tedious interaction I've described problematic.

There's a simple solution that eliminates the tedium: *for each Ghost, show the card it turns into at the start of combat*. This means that everyone will now have this information. No tedium for the players who would have taken the time to work it out, extra information for the people who wouldn't have. And the "which card is the Ghost" dynamic still works, you just know which set of cards it's going to be in.

I recognize that this is something where there could be a view that "it's intended to be fun to play without thinking about which card is the Ghost, we don't want people to know, we shouldn't encourage people to make the no-fun play even if might be optimal". But we should recognize that there's a broader spectrum of people playing Slay the Spire with more than one idea of what's fun, and leaving things the way they are will create anti-fun for a portion of players.

Full disclosure: I'm one of those players, and this is the type of thing that would probably make me turn off a pack. If there's a card that is unpickable because it makes the run not fun to play, I generally don't want to play with the pack that has that card.

modargo:
Balance feedback for The Prism: I'm looking at the option of taking this from the act 1 boss rewards right now, and it feels a bit hard to take. Taking it puts 3 attacks into my hand/draw/discard that are worse than Strike+ (a card you want to remove from your deck), and in any combat where I manage to play all 3 and then draw The Prism itself aftewards, I could have used the turns/energy/cards to do more than what I've accomplished by playing The Prism.

For an example using cards in the deck for the current run where I have this choice, I could use the same 6 energy and 4 card plays (across two deck cycles) to play a Carnage+ twice and a Synergize+ twice. That comes up to 74 damage (18 of it aoe) and 18 block. Pretty comparable to the 108 damage that 3 Shard+s and Prism+ would give, and much more frontloaded, which is very important in hallway fights.

Colorless synergy cards can improve the speed of this process or give you more reasons to play Shards, but the first time you play The Prism still isn't very cost effective, and playing it more than once will normally only happen in boss fights and a few elite fights. What that means is instead of taking The Prism, you'll want to take cards that will win those fights without slowing down your other fights (which a decent number of other cards will do). Imagine that instead of doing the whole Prism thing, you just played a Footwork+, a Noxious Fumes+, and some other poison and block cards.

Really up to you if you want to improve this card or just leave it as is. One way to improve the card would be to make the Shards good enough that it didn't feel like you were making your deck worse by having them. They could deal 10 damage instead of 8, or also draw a card, or cost 0 instead of 1 (and deal a bit less damage).

src/main/java/thePackmaster/cards/colorlesspack/GolfBall.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public void applyPowers() {
+        this.baseDamage = BLOCK_AMT_LOST;
+        super.applyPowers();
+        this.rawDescription = cardStrings.DESCRIPTION;
+        this.rawDescription = this.rawDescription + cardStrings.EXTENDED_DESCRIPTION[0];
+        this.initializeDescription();
+    }
+
+    public void onMoveToDiscard() {
```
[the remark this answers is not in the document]
DarkVexon: I figured for consistency I should just do what basegame does... hmm

src/main/resources/anniv5Resources/summaries/ratings.txt
DarkVexon: Done!

src/main/java/thePackmaster/cards/monsterhunterpack/AwakenedRitualDagger.java
```diff
@@ -18,7 +18,7 @@ public class AwakenedRitualDagger extends AbstractMonsterHunterCard {
     public static final int UPG_DAMAGE = 10;
 
     public AwakenedRitualDagger() {
-        super(ID, 1, CardType.ATTACK, CardRarity.SPECIAL, CardTarget.ENEMY);
+        super(ID, 1, CardType.ATTACK, CardRarity.SPECIAL, CardTarget.ENEMY, CardColor.COLORLESS);
```
modargo, line 21: For all these Monster Hunter cards, is there are consideration to leaving them as the Packmaster color? They feel different to me than all the other tokens, since they get added to your deck permanently instead of being generated in combat. 

I don't really have a strong opinion either way, but while making everything generated during combat colorless makes sense, these are a bit different. They're almost like transformations of the Carving Knife that was used to kill the enemy and gain the new card.

erasels: I'd prefer them not being colorless for the reason you stated. I feel like they're an integral part of the pack and not really a token that you generate often

PR #539, opened 2023-03-24
Calamity pack

modargo:
After additional playtesting, I got this pack to a place I'm happy with. I think the only change from the version I previewed was to Cobra's Gift -- it upgrades to not exhaust, which is a fairly good option (there aren't that many ways to get non-exhausting energy-positive cards). I've found it to be fun to play with, particular in combination with any other pack that involves debuffs. Frostbite also gives it fairly good synergy with any pack that has good attacks.

As with my previous post-release pack, this needs a frame and a hat, and I'll follow up with those after this is merged.

No rush on merging this if you want to get a bugfix release out before releasing more packs.

calamity pack cards
calamity pack cards upgraded


src/main/java/thePackmaster/actions/summonspack/JinxLoseHPAction.java
```diff
@@ -25,7 +25,7 @@ public JinxLoseHPAction(AbstractCreature target, AbstractCreature source, int am
 
     public void update() {
         if (duration == DURATION && target.currentHealth > 0 && !quiet)
-            CardCrawlGame.sound.play(SpireAnniversary5Mod.EVIL_KEY, 0.1f);
+            CardCrawlGame.sound.playAV(SpireAnniversary5Mod.EVIL_KEY, 0.1f, 0.5f);
```
modargo, line 28: I made this quieter since the Calamity pack can apply it with a common, and I found it a bit too loud when it was going on regularly due to more debuffs being applied.

src/main/java/thePackmaster/vfx/calamitypack/EssenceGraspEffect.java
```diff
@@ -0,0 +1,44 @@
+package thePackmaster.vfx.calamitypack;
+
+import com.badlogic.gdx.Gdx;
+import com.badlogic.gdx.graphics.g2d.SpriteBatch;
+import com.badlogic.gdx.math.MathUtils;
+import com.megacrit.cardcrawl.core.Settings;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.vfx.AbstractGameEffect;
+
+public class EssenceGraspEffect extends AbstractGameEffect {
```
modargo, line 10: These effects are edited versions of what Hemokinesis does -- black tendrils that reach out to the enemy.

src/main/resources/anniv5Resources/summaries/ratings.txt
```diff
@@ -12,6 +12,7 @@ BitingColdPack	4	2	2	3	3
 BlacksmithsPack	3	3	4	4	4
 BoardGamePack	4	3	2	4	2
 BulwarkPack	2	5	1	4	4
+CalamityPack	4	2	2	3	4
```
modargo, line 15: From my experience playtesting. The pack is good offensively, has some defense and support (but a bit more niche or at higher rarities), is fine on frontload (the common attacks are pretty good), and provides strong scaling options.

modargo:
What do you think of this for a hat? I think I'd call it "On Fire" or just "Fire". Let me know if you think it's better to let AlexMdle make something.
image
CalamityPackHat

erasels:
Amusing hat, alch added support for hats that just spawn vfx, we have a fire particle in base game you could use if that's something you're interested in. Otherwise, I think the hat is fine.

modargo:
> Amusing hat, alch added support for hats that just spawn vfx, we have a fire particle in base game you could use if that's something you're interested in. Otherwise, I think the hat is fine.

Oh interesting, if I'd known about support for hats that spawn vfx beforehand I might have tried to make something using that. I'm happy enough with this hat that I don't feel the need to tinker further, though.

src/main/java/thePackmaster/powers/calamitypack/GrowingFlamesPower.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public GrowingFlamesPower(AbstractCreature owner, int amount) {
+        super(POWER_ID, NAME, PowerType.BUFF,false, owner, amount);
+    }
+
+    @Override
+    public void atStartOfTurnPostDraw() {
+        if (!AbstractDungeon.getMonsters().areMonstersBasicallyDead()) {
+            this.flash();
+            Wiz.forAllMonstersLiving(m -> this.addToBot(new ApplyPowerAction(m, this.owner, new IgnitePower(m, this.amount))));
```
erasels, line 27: There is an AllEnemyApplyPowerAction, that might help here because this will be slow

src/main/resources/anniv5Resources/localization/eng/calamitypack/Cardstrings.json
```diff
@@ -0,0 +1,50 @@
...
+    "NAME": "Doomtouch",
+    "DESCRIPTION": "Deal !D! damage. NL Apply !M! ${ModID}:Jinx."
+  },
+  "${ModID}:BindingChains": {
+    "NAME": "Binding Chains",
+    "DESCRIPTION": "Gain !B! Block. NL Enemy loses !M! Strength this turn. NL Exhaust."
+  },
+  "${ModID}:FreezingVenom": {
+    "NAME": "Freezing Venom",
+    "DESCRIPTION": "Deal !D! damage to ALL enemies. NL Apply !M! ${ModID}:Frostbite and Poison to ALL enemies."
```
erasels, line 12: Bit of a mismatch between the description and the actual effect.
Would recommend doing the normal AoE damage here and then using the AllEnemyApplyPowerAction to apply the powers. This would also speed up the distribution.

Otherwise, I'd ask you to reword this to say each enemy, not ALL enemies.

erasels (changes requested):

PR #553, opened 2023-03-29
- goddess of explosions balance tweaks

AlexMdle:


src/main/resources/anniv5Resources/localization/eng/goddessofexplosionspack/Cardstrings.json
```diff
@@ -27,7 +27,7 @@
   },
   "${ModID}:MoreExplosions": {
     "NAME": "More Explosions!",
-    "DESCRIPTION": "Shuffle !M! random Powers into your draw pile."
+    "DESCRIPTION": "Draw !M! cards. NL Shuffle !M! random Powers into your draw pile."
```
modargo, line 30: So this will now be a 0-cost draw 2 (3 when upgraded)? That seems a bit too strong — drawing multiple cards for 0 normally comes with a downside (eg Overclock, Battle Trance) or with high rarity and exhausting (eg Master of Strategy, Scrawl if you want to count that).

Maybe have this draw 1 card instead. That should be a good buff to the card still.

AlexMdle: It was a result of discussion in modding. Turned out shuffling powers into the draw pile is actually really, really bad. Initially wanted 1(2) draw, but sandtag convinced me that would still be bad - 1 draw would not make up for the card itself taking up a draw, while still trashing the draw pile on use. Its effectively an alt. overclock now, except instead of 1 burn its 2(3) powers.

erasels: I don't see a world where random powers are comparable to a negative status like burn. Especially in a pack that actively synergizes with playing them

AlexMdle: Closer to a wound fairly often, but also 2(3) times as many as the comparison card.

modargo: If we view shuffling powers into the draw pile as a downside, my opinion that we should redesign the card. For example, have it add the powers to your hand instead. Given that we have cards like Grab Bag (which adds 4 cards different color cards to your hand for 1 energy, upgrade to 0 energy), that's not inherently crazy. We would need to work out whether 2 (3) is reasonable for powers -- it's not entirely clear, since comparing them to different color cards, powers from your packs are more likely to have relevant scaling, but less likely to have immediate damage or block. But it seems like a reasonable redesign to me with 2 (3) as a starting point.

If we view shuffling powers into the draw pile is an upside that requires some conditions and setup to activate (which is my point of view since this pack has decent power synergy), giving this draw 1 is reasonable, and giving it draw 2 (3) is too much.

AlexMdle: That's fair, let's see how it performs with 1(2) draw first.

PR #556, opened 2023-04-01
s rk e buff

Alchyr:
More Distortion, less Damage


src/main/java/thePackmaster/cards/distortionpack/S_r_ke.java
```diff
@@ -30,6 +30,6 @@ public void use(AbstractPlayer p, AbstractMonster m) {
     }
 
     public void upp() {
-        upgradeDamage(3);
+        upgradeMagicNumber(1);
```
modargo, line 33: If you want I think you could have the damage go up by 1 too. It's a common pattern with upgrades for cards like this (common attacks that deal Strike-level damage and give a small buff, debuff, or other benefit). Two examples are Pommel Strike and Poison Stab from the base game.

I suggest this for reasons of design aesthetics, not balance -- there's something about _both_ numbers turning green in the upgrade preview that just feels good. In terms of balance I think it's fine either way (1 damage on a common attack like this isn't going to make or break it unless you start getting under or over the typical range).

PR #560, opened 2023-04-02
Artificer pack

Pandemonium14:
Full transparency about the hand ordering, because we had a discussion about it :
While a hand select screen is up, the hand order does not change. Cards added to the hand as a result of the action with a AddTempCardInHand action are put to the right.
The one wrinkle it has is with actions that open a hand select screen and say, add a modifier to the selected card, because then, after clicking Done, the card is put to the right of the hand, not in its original place. This is consistent still, so I'm personally okay with it, but maybe you're not?
I don't immediately see how to change that behavior, but if it bothers you, just say so and I'll try to fix the issue.


src/main/resources/anniv5Resources/localization/eng/artificerpack/Cardstrings.json
```diff
@@ -0,0 +1,42 @@
...
+    "NAME": "Gravity Slam",
+    "DESCRIPTION": "Deal !D! damage. NL Exhaust adjacent cards."
+  },
+  "${ModID}:GrantSwiftness": {
+  "NAME": "Grant Swiftness",
+  "DESCRIPTION": "Draw !M! cards. NL ${ModID}:Boost adjacent cards: NL \"Draw 1 card\""
+  },
+  "${ModID}:WhiteSteelCharm": {
+    "NAME": "White Steel Charm",
+    "DESCRIPTION": "Unplayable. NL When you play an adjacent card, a random enemy takes !M! damage."
```
erasels, line 24: Hmm, wouldn't it be Whenever here?

erasels (changes requested):

Pandemonium14:
Should be done? I'm not sure what else to do about Boost's wording tbh

src/main/java/thePackmaster/cardmodifiers/artificerpack/BlockEnchantModifier.java
```diff
@@ -25,13 +25,14 @@ public BlockEnchantModifier(int amount) {
...
-                    ((BlockEnchantModifier) m).amount += amount;
-                    amount = 0;
-                }
+            if (m instanceof BlockEnchantModifier && m != this) {
+                ((BlockEnchantModifier) m).amount += amount;
+                amount = 0;
             }
         }
+        if (amount == 0) {
+            CardModifierManager.removeSpecificModifier(card, this, true);
```
erasels, line 34: Might want to make this an anonymous action. This could lead to a conc mod exception

erasels (changes requested):

Pandemonium14:
Done!

PR #565, opened 2023-04-05
Overwhelming pack

Alchyr:
art in progress

Sheer Force - concerns me a bit, might need adjustment
Giga Drill being Kinda Sorta Weaker Meal Prep but Not Really? Should be fine, I think


erasels:
Haven't looked at the code yet but some of the cards here seem kind of problematic to me:
Lariat and Opening Strike seem incredibly strong. Pretty above the curve power-wise.
Withstand and Giga Drill are ver ysimilar to cards that are already in packs (Reinforce Body, kinda Mind Maze and the card from Energy & Echo pack)
Sheer Effort seems problematic balance wise, spend 2 energy to play 2 cards regardless of cost. This will be unlikely to ever be low-value.
Overcome seems like it requires a good amount of support to even work, we only have like 3 packs that support it? Might be problematic in the opposite way.
Further Beyond needs a decent amount of support as well, it's a double hoop to jump through (no energy and somehow being played). The payoff doesn't seem big enough for that.
I don't think Make Room interacts well within the pack. We have a decent amount of skims but we don't really do crazy draw shenanigans outside of Flow pack. I'd rather this card be reworked.
I **really** do not like Final Form. I would ask you to remove the card and make a new one. Form cards are problematic at the best of times, being able to easily play them is no bueno. 

This pack has a theme going for it that I really enjoy and some of the effects sound very fun but it's important to keep in mind that that there's 60 other cards in the draft that you are just as likely to see. I think the pack probably needs another way of cost reduction/negation/whatever that helps with the payoff cards its trying to add.
Oh and the art is great!

Alchyr:
note of some changes - 
Giga Drill - 
14 damage, top 1 [2] cards of draw pile cost 0 until played
Withstand (not 100% confirmed) - 
_Probably_ 1 cost, 5 block, random card in hand -1 cost for turn. Would also be moved to common if so. More block than a deflect but the "energy" saved is random, and then it adds more accessibility to synergy stuff.

Rummage helps a lot with overcome being pickable, and you can get more value by finding more stuff for it from the pack/random other things like mummified hand, snecko eye, any other card reduction or card doubling

Sheer Effort - Yeah, it's tricky. I think it'd just be bad at 3 cost, but I feel like having it at 2 cost is a bit too free. Might need a rework of some kind.
Lariat - Might need a damage nerf, idk. It has Very High Potential but a lot of the time you'll just save 1 energy.
Opening Strike - Definitely a Card. One energy discount means you aren't cheating any other expensive cards; would compare to bullet time. No draw restriction and has some damage attached, but not quite as powerful of an effect. (Also is uncommon, though.) Definitely stronger the later in the run you get.
Further Beyond - It's kinda more like 1.5 hoops; if you get it to 0 cost getting to 0 energy isn't usually a problem
Alternatively you cheat it out with some other card play effect while getting yourself to 0 energy. That said, it's being adjusted to base 1 cost, which makes it easier to play without the upgrade.
Make Room is fine; if you have more draw it can do a lot, and if you don't then it's just a Decent Draw card
Final Form - it's not *that* easy, but the card is definitely risky to have in the pack. The cards you get from it will still cost energy, but I'm not sure if it's too Strong or Too Hard to use.

modargo:
I gave feedback for this pack in Discord and I see the discussion has now moved here, so I wanted to briefly weigh in. Gk has covered a lot of similar points, which is a pretty clear sign of where changes are needed. I'll stick to amplifying a few things that still seem to be in question and bringing up one card that wasn't really mentioned:
* Sheer Effort: Like I mentioned in Discord, this is effectively energy generation equal to the two most expensive cards in your hand as a baseline. That's so hard to balance that I don't know if you can with the current design. You'd need to add an energy limit for the cards, e.g. "Play two cards that cost 2 or less". At that point maybe you just rework the card entirely.
* Final Form: This just has too many issues; the pack will be much better off with a rare that's more consistent than this can ever be.
* Further Beyond: I think the pack could find a better rare that isn't so dependent on other cards from the pack to enable it, but I guess it's not as much of a problem as Final Form if the pack ends up with enough support for it.
* Approach: I'm still skeptical that this will play well. Drawing until you get a 1-cost seems like it will be very inconsistent, even with a bit of Scry. And sometimes it will be broken if you manage to stack your draw pile just right. Not the best profile for balance.

Given how much of the pack is in question, it might be worth taking a step back and thinking about the pack as a whole. Which cards and concepts are really essential? What are the core ideas that resonate and should be emphasized and built upon? What are the extraneous or difficult to balance parts that aren't worth including? How can support and payoff make a balanced mix that also interacts with the other packs in a typical run? I would expect that to naturally lead to reworking/replacing a number of cards as part of tightening up the overall design of the pack.

That may also be a faster way to get this pack to a good place than trying to debate the feedback for every card. Obviously it's up to you how you want to work on this, but it strikes me that the card-by-card approach combined with the way this feels like a game of telephone (since you aren't the original designer of this pack) could end up being pretty inefficient.

erasels:
Art hasn't been touched ina  while as well, so I figure this pack is probably dead?

Alchyr:
aight here's current version of ideas

image
image

Alchyr:
all cards are coded besides Final Form

src/main/java/thePackmaster/cards/overwhelmingpack/Withstand.java
```diff
@@ -1,26 +1,42 @@
...
 
     public void use(AbstractPlayer p, AbstractMonster m) {
-        this.addToBot(new ReinforcedBodyAction(p, this.block, this.freeToPlayOnce, this.energyOnUse));
+        int blck = this.block, threshold = this.magicNumber;
+        addToBot(new AbstractGameAction() {
+            @Override
+            public void update() {
+                this.isDone = true;
+                for (AbstractCard c : p.hand.group) {
+                    if (Wiz.getLogicalCardCost(c) <= threshold) {
```
erasels, line 31: Unplayable cards are treated as 0 cost for that method, up to you if you want that.

erasels (commented):
Cursory glance didn't show any mistakes, please do test the cards when you#re done though. Will sit down and do art for it soon

src/main/java/thePackmaster/cards/overwhelmingpack/Withstand.java
Alchyr: (I do)

Alchyr:
made final form work, tested stuff, added some power images and set pack image
those 4 cards still need art, and then I'll probably decide art for the Overcome power icon after that

src/main/resources/anniv5Resources/localization/eng/overwhelmingpack/Powerstrings.json
```diff
@@ -7,6 +7,14 @@
       " #ySkills you play."
     ]
   },
+  "${ModID}:ActuallyFinalFormPower": {
+    "NAME": "Final Form",
+    "DESCRIPTIONS": [
+      "At the end of your turn, play the highest card card in your hand.",
```
erasels, line 13:
```suggestion
      "At the end of your turn, play the highest cost card card in your hand.",
```

src/main/resources/anniv5Resources/localization/eng/overwhelmingpack/Powerstrings.json
```diff
@@ -7,6 +7,14 @@
...
+      "At the end of your turn, play the highest card card in your hand #b",
```
erasels, line 14:
```suggestion
      "At the end of your turn, play the highest cost card card in your hand #b",
```

erasels (changes requested):
Oh and please don#t forget to add the hat strings, and pack ratings. They can be placeholder, just so we don#t froget

src/main/resources/anniv5Resources/localization/eng/overwhelmingpack/Powerstrings.json
Alchyr: 👍

src/main/resources/anniv5Resources/localization/eng/overwhelmingpack/Powerstrings.json
Alchyr: 👍

Alchyr:
for now,
offense 4
defense 2
support 4
frontload 3
scaling 4

hat maybe just make packmaster's hair go super saiyan
it's already the right color

erasels:
Alrighty, sent you the last images via discord. Those + the missing entry in the pack tags (even if it's just none) and we're good to go!

Alchyr:
aight added stuff

erasels:
Thank you for the contribution :D

PR #569, opened 2023-04-06
Shocking

Jedi515:


erasels:
Art is still in the works I presume?

src/main/java/thePackmaster/actions/lockonpack/LightsOutAction.java
```diff
@@ -0,0 +1,47 @@
...
+                if (upgraded)
+                {
+                    addToTop(new EvokeSpecificOrbAction(orbs.get(i)));
+                    addToBot(new ChannelAction(new Dark()));
+                }
+                else
+                {
+                    Dark d = new Dark();
+                    orbs.set(i, d);
+                    d.setSlot(i, Wiz.adp().maxOrbs);
```
erasels, line 40: This wouldn't trigger orb channeling hooks. Kinda suspect but since you worded it as replace it's probably fine

src/main/java/thePackmaster/cardmodifiers/lockonpack/GlockOnModifier.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class GlockOnModifier extends AbstractCardModifier {
+
+    @Override
+    public float modifyDamage(float damage, DamageInfo.DamageType type, AbstractCard card, AbstractMonster target) {
+        return damage + Wiz.getLogicalPowerAmount(AbstractDungeon.player, FocusPower.POWER_ID);
+    }
+    @Override
+    public float modifyDamageFinal(float damage, DamageInfo.DamageType type, AbstractCard card, AbstractMonster target) {
+        if (target == null) return super.modifyDamage(damage, type, card, target);
```
erasels, line 21: Need an explanation for what's happening here. Wouldn't this just not work for AoE attacks? 
What's the leftover var below for? Some weird form of casting?

src/main/java/thePackmaster/cards/lockonpack/AC.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public AC() {
+        super(ID, 2, CardType.SKILL, CardRarity.COMMON, CardTarget.SELF);
+        baseBlock = block = 10;
+        cardsToPreview = new DC();
+    }
+
+    @Override
+    public void upp() {
+        upgradeBlock(4);
```
erasels, line 25: Upgrade the preview card as well here

src/main/java/thePackmaster/cards/lockonpack/TunnelVision.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    @Override
+    public void upp() {
+        upgradeMagicNumber(2);
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        addToBot(new ApplyPowerAction(p, p, new TunnelVisionPower(p)));
```
erasels, line 26: Magic number isn't used anywhere? Don't see Lock on being applied here

src/main/java/thePackmaster/cards/lockonpack/special/DC.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Override
+    public void upp() {
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        Lightning orb = new Lightning();
+        addToBot(new ChannelAction(orb));
+        if (upgraded) orb.onEndOfTurn();
```
erasels, line 33: Seems problematic to do this before the orb is even channeled. Might be better in an anonymous action just to be sure

src/main/java/thePackmaster/cards/lockonpack/special/DC.java
```diff
@@ -0,0 +1,36 @@
...
+import thePackmaster.util.Wiz;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class DC extends AbstractLockonCard {
+
+    public final static String ID = makeID(DC.class.getSimpleName());
+
+    public DC() {
+        super(ID, 1, CardType.SKILL, CardRarity.SPECIAL, CardTarget.NONE);
```
erasels, line 20: Might be worth making this token colorless

src/main/java/thePackmaster/patches/lockonpack/ActualLockOnPatch.java
```diff
@@ -0,0 +1,40 @@
...
+
+@SpirePatch2(clz = Lightning.class, method = "triggerPassiveEffect")
+@SpirePatch2(clz = LightningOrbPassiveAction.class, method = "update")
+@SpirePatch2(clz = LightningOrbEvokeAction.class, method = "update")
+public class ActualLockOnPatch {
+    public static AbstractMonster getLockedMonster(AbstractMonster originalTarget)
+    {
+        if (originalTarget.hasPower(LightningRodPower.POWER_ID)) return originalTarget;
+
+        ArrayList<AbstractMonster> lockedList = AbstractDungeon.getMonsters().monsters.stream().
```
erasels, line 25: This also gets dead monsters, doesn't it? We have a method in Wiz to get the living enemies

src/main/java/thePackmaster/patches/lockonpack/ActualLockOnPatch.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.ArrayList;
+import java.util.stream.Collectors;
+
+@SpirePatch2(clz = Lightning.class, method = "triggerPassiveEffect")
+@SpirePatch2(clz = LightningOrbPassiveAction.class, method = "update")
+@SpirePatch2(clz = LightningOrbEvokeAction.class, method = "update")
+public class ActualLockOnPatch {
+    public static AbstractMonster getLockedMonster(AbstractMonster originalTarget)
+    {
+        if (originalTarget.hasPower(LightningRodPower.POWER_ID)) return originalTarget;
```
erasels, line 23: Might be neat to add some logic to hit all enemies with the power instead of only the first

src/main/java/thePackmaster/patches/lockonpack/AdditiveLockOnPatch.java
```diff
@@ -0,0 +1,60 @@
...
+import thePackmaster.powers.lockonpack.TunnelVisionPower;
+import thePackmaster.util.Wiz;
+
+public class AdditiveLockOnPatch {
+
+    public static int applyMultiLockOn(AbstractMonster m, int damage)
+    {
+        float damageModifier = 1.5F
+                + 0.25F
+                * (Wiz.getLogicalPowerAmount(m, LockOnPower.POWER_ID) - 1);
```
erasels, line 22: Does this work correctly if you ahve no lock-on currently? This probably won't be called then but just making sure you thought about that.

Doesn't this need a check for Tunnelvision being active?

src/main/resources/anniv5Resources/localization/eng/lockonpack/Cardstrings.json
```diff
@@ -0,0 +1,43 @@
...
+    "DESCRIPTION": "{@@}Evoke your rightmost Lightning !M! times hitting ALL enemies{!Upgrades!|0= NL Exhaust.|>0=.}"
+  },
+  "${ModID}:LightsOut": {
+    "NAME": "Lights Out",
+    "DESCRIPTION": "Replace your Lightning with Dark. NL Exhaust.",
+    "UPGRADE_DESCRIPTION": "Evoke your Lightning and Channel Dark for each Lightning Evoked. NL Exhaust."
+  },
+  "${ModID}:GlockOn": {
+    "NAME": "Glock-On",
+    "DESCRIPTION": "Deal !D! damage. NL Damage is also affected by Focus and Lock-On."
```
erasels, line 25:
```suggestion
    "DESCRIPTION": "Deal !D! damage. NL Affected by Focus and Lock-On."
```
Could be less wordy

erasels (changes requested):

src/main/java/thePackmaster/actions/lockonpack/LightsOutAction.java
Jedi515: It's intended :tm: (would otherwise require more jank)

src/main/java/thePackmaster/cardmodifiers/lockonpack/GlockOnModifier.java
Jedi515: I could either do maths in card itself with ap/ccd, or make an inherent cardmod. This is the latter. Only intended to be used on one card, but, well, if you wish.
The leftover var is because normal damage is calculated with floats, but lock-on only runs ints. Will update a bit too.

src/main/java/thePackmaster/patches/lockonpack/AdditiveLockOnPatch.java
Jedi515: It's only getting called if lock-on is on... but it does need tunnelvision check.

Jedi515 (commented):
Code solutions underway, discussion "solutions" are all here.

src/main/java/thePackmaster/cards/lockonpack/TunnelVision.java
```diff
@@ -23,6 +25,10 @@ public void upp() {
 
     @Override
     public void use(AbstractPlayer p, AbstractMonster m) {
+        if (magicNumber > 0)
+        {
+            for (AbstractMonster mon : Wiz.getEnemies()) addToBot(new ApplyPowerAction(mon, p, new LockOnPower(mon, magicNumber)));
```
erasels, line 30: We have an action to apply powers to all enemies. It's a lot faster than doing ti one by one, please use that

erasels (changes requested):

erasels:
Okay, other than the possible wording change and making DC colorless, this seems ready to merge

PR #570, opened 2023-04-06
Three Pack Updates - Gates 2 and 3 and Eldritch Horror

mikemayhemdev:
Gate 2: Conjuration Updates

Magic keyword removed entirely.  This was too self-serving and was not able to be properly integrated into enough PM packs to make it work.  The Conjuration pack is now solely structured around generating tokens and playing lots of cards.

Flaming Grasp: Now grants a Flaming Grasp next turn instead of on next Attack.

Flame Pillar: Now just Ignites for 3.

Conjure Barrage: Magic Missiles now only affect themselves.

Crusher: Reduced cost to 5 (from 6)

Haymaker: Damage per card played increased to 3.  Upgrade now increases base damage by 2 in addition to adding 1 damage for each card played.

Arsenal Gear: Now reduces the generated card's cost by 1 instead of reducing to 0, and does not add Exhaust.

Mana Shield's Channel: Now reduces the cost of a random card in hand by 1 this turn. (instead of only hitting Magic cards)

Daggerstorm reworked, now reads: Whenever you generate a card other than Shiv, create and play a Shiv on a random target.

Wall of Fire: Upgrade now increases damage by 2 and upgrades the 3 Flame Pillars, instead of adding a 4th hit.  Upgraded Flame Pillars apply 2 more Ignite each.

Gate 3: Life & Death Pack Update

Minor update to this pack to primarily give it more self-synergy and the ability to kick off more combos, mostly revolving around the new Mantis Strike.

Mantis Strike reworked, now reads: "Deal 4(6) damage to a random enemy. Repeat each time you fully block damage this turn.".  This should serve as a bread & butter common for this deck without being self-damaging, while still being useful for drafts that don't self-damage.

Corset: Changed to "Gain 10 (13) Block. Add a Jam to your hand.".  This gives the card a more dynamic playstyle - you can choose to play it like a Shrug it Off that triggers Mantis Strike/Ouroborous, or play it like a better Leap but at the expense of the Jam staying in your deck.

Self Wound - Now provides 4(6) Temporary HP instead of healing and no longer Exhausts.

Minniegun - Jam now goes to the draw pile instead of discard pile.

Inferno - Downside changed to 'Add 2 Burns to your hand'. This gives a bit better synergy with Mantis Strike & Ouroborous.

Spreading Spores - Upgrade changed to "shuffle a copy of this into your draw pile". (instead of +2 Ablation).

Ouroborous - Is now Rare.  This was a bit too easy access to infinite strength scaling for an Uncommon slot.

Restoration Detonation - Heal increased to 10 (from 8).

Eldritch Horror Pack Update

Large update to this one.  This pack needed a stronger identity, it was trying to do too many things and tie into too many new PM keywords in a somewhat arbitrary manner.  The new mechanic Sanity is now the primary driver of the Eldrtich Horror pack, which can let a carefully constructed draft cheat out high cost cards.

New Mechanic - Sanity: While you have negative Sanity, lose 1 Sanity at the start of each turn. For every 5 sanity lost, add a Lunacy to your hand.

New Token Card - Lunacy: Costs 0, and reduces the cost of a random card in hand to 0 until played. Also causes you to lose 1 HP and 1 Sanity.

Beyond the Stars - Now reads 'Lose 1 Sanity for each enemy hit", damage bumped up to 7.  Does not generate Madness.

Nameless Mist -  Block increased to 5(8). Secondary effect is now Flurry: Gain 1 Temporary HP.

All Seeing (now named Eldritch Pact) reworked, now reads: "At the start of each turn, lose 2 Sanity and gain 2 Block.".  When upgraded, changes the secondary effect to "Whenever you lose Sanity, gain 2 Block."

Flailing Tendril - Now causes you to lose 3 Sanity, damage increased by 1.  Secondary effect reworked to "If this costs 0, hits twice."

Gaze the Void reworked, costs 1, now reads: "Deal 5(8) damage to random enemy. Repeat for each Lunacy you've played this combat."

Page of the Dead - Rework completely - "Lose 3 Sanity. Add a random Attack that costs 2 to your hand."

Familial Curse - Now applies 1(2) Weak to ALL enemies and grants 10 Block. Causes you to lose 4 Sanity.

R'lyeh Ftaghn Reworked, now reads: Go to sleep (End your turn). Next turn, gain 15(20) Temporary HP, gain 3 E, and enter Nightmare.

While in Nightmare Stance, lose 1 sanity when you play a card, and draw 2 cards when you play a Madness.
Exit Nightmare Stance at the end of the turn.

Bonus Changes:

Frost Pack - Extended Stall now reduces the cost of the card hit by 1.


src/main/java/thePackmaster/cards/cthulhupack/PageOfTheDead.java
```diff
@@ -14,17 +13,20 @@ public class PageOfTheDead extends AbstractCthulhuCard {
...
 
-        cardsToPreview = new Necronomicurse();
     }
 
     public void use(AbstractPlayer p, AbstractMonster m) {
-        if (!upgraded) addToBot(new MakeTempCardInHandAction(new Necronomicurse()));
-        Wiz.applyToSelf(new PageOfTheDeadPower(p, 1));
-
+        loseSanity(magicNumber);
+        AbstractCard strike = Wiz.returnTrulyRandomPrediCardInCombat(card -> card.type == CardType.ATTACK
```
modargo, line 23: It's possible to have a set of packs with no 2-cost attacks. Eldritch Horror itself doesn't have any, and it isn't too hard to find 6 other packs without 2-cost attacks. In fact, from a quick look through things, there's a _lot_ of packs without 2-cost attacks, so this is pretty likely to happen if you play enough. I ran into a crash for something similar to this yesterday -- I had a set of packs with only 1 rare attack, and I did something that assumed there would be at least 2 rare attacks.

If there are no 2-cost attacks in your packs, this will return null and crash. We should have some sort of handling for that -- I'm not sure what option is best. Creating nothing is simplest but feels pretty bad if you take this card without realizing you don't have any 2-cost attacks. An alternative is to fall back to creating a different card (like Hand of Greed, which I think is the only colorless 2-cost attack) or fall back to looking across all base game 2-cost attacks.

A corollary to this is that there are sets of packs that only have one 2-cost attack, where playing this will give you the same card every time. Don't know if you need to do anything about that, but figured I'd mention it.

Also I'd suggest renaming this to `attack` instead of `strike`.

modargo (changes requested):
Took a bit to get to this but I've gone through and added comments! Thanks for writing out the summary of changes, it was helpful to reference while looking back and forth between the PR and the cards in game.

I like all the changes to the Gate packs (with one exception that I commented on; Self Wound). I took a bit to think about the balance of each card and left some thoughts where I though things might be off -- Mana Shield and Inferno are the two cards I had some concern with. Things look good overall though, commented on a few minor issues and whatnot but nothing big.

The redesign of Eldritch Horror pack is interesting. I'm not sure what how it will play, but the idea makes sense. The only big design issue that struck me was the two unchanged rares feeling out of place given the new design for the pack -- changing them to be more on theme would be a big positive IMO. Also did a bit of testing and commented on some small issues I found. Bit harder to tell what the balance will be like.

There's a lot in this PR so it will probably make sense for me or Gk to take another pass for anything that the first review missed, after changes based on my comments. Overall these redesigns look great so I'm excited to get them in before the next release! Not sure exactly when that will be, but I'd assume this upcoming weekend or the weekend after that, depending on Gk's schedule and what the outstanding PRs look like.

src/main/java/thePackmaster/cards/cthulhupack/PageOfTheDead.java
modargo: Falling back to making a Rummage is another option.

src/main/java/thePackmaster/powers/cthulhupack/SanityPower.java
```diff
@@ -0,0 +1,68 @@
...
+            AbstractDungeon.onModifyPower();
+        }
+    }
+
+    @Override
+    public void stackPower(int stackAmount) {
+        this.fontScale = 8.0F;
+        this.amount += stackAmount;
+        while (amount <= -5) {
+            amount += 5;
```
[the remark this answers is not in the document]
mikemayhemdev: The only reason I left it here is because when in use with the pack in general, you'll more often than not go to 0, and then reduce it more before the turn is done by playing the Lunacy you get.  This setup makes the power icon stay in place in that case instead of having it shift to the right again.  A minor convenience but will get rid of it if it's a problem.

src/main/java/thePackmaster/cards/cthulhupack/PageOfTheDead.java
mikemayhemdev: This function does not go through the pack list, it pulls from all game cards.  Just tested to verify, got Brainwave while not using the Psychic pack.

src/main/java/thePackmaster/cards/dimensiongatepack3/SelfWound.java
```diff
@@ -16,15 +17,13 @@ public class SelfWound extends AbstractDimensionalCardTowerTactics {
...
-        baseMagicNumber = magicNumber = 3;
-        tags.add(CardTags.HEALING);
+        baseMagicNumber = magicNumber = 4;
     }
 
 
     public void use(AbstractPlayer p, AbstractMonster m) {
         addToBot(new SelfDamageAction(new DamageInfo(p, 3, DamageInfo.DamageType.THORNS), AbstractGameAction.AttackEffect.FIRE));
-        Wiz.atb(new HealAction(p, p, magicNumber));
+        Wiz.atb(new AddTemporaryHPAction(p,p, magicNumber));
```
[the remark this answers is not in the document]
mikemayhemdev: I think I'm going to let this one as is, I think there's real power to converting block to permanent block for 0 on turns where you don't have any better action to take.  Mostly though this is important fuel to have a 0-cost self-damage trigger for Mantis Strike.

src/main/java/thePackmaster/cards/dimensiongatepack3/SpreadingSpores.java
```diff
@@ -35,11 +35,11 @@ public List<TooltipInfo> getCustomTooltips() {
     }
 
     public void use(AbstractPlayer p, AbstractMonster m) {
-        Wiz.applyToSelf(new ThornsPower(p, secondMagic));
+        Wiz.applyToSelf(new ThornsPower(p, magicNumber));
         Wiz.applyToSelf(new TempHPRegenPower(p, magicNumber));
+        if (upgraded) Wiz.atb(new MakeTempCardInDrawPileAction(this.makeStatEquivalentCopy(), 1, true, true));
```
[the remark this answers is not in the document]
mikemayhemdev: I'm game to try that, was just a bit concerned on power level.

src/main/resources/anniv5Resources/localization/eng/cthulhupack/Cardstrings.json
```diff
@@ -1,46 +1,51 @@
...
+    "EXTENDED_DESCRIPTION": [" NL (Hits ", " times.)"]
   },
   "${ModID}:AllSeeing": {
-    "NAME": "All Seeing",
-    "DESCRIPTION": "At the end of each turn, Scry !M! and take 3 damage."
+    "NAME": "Eldritch Pact",
+    "DESCRIPTION": "At the start of each turn, lose !M! ${ModID}:Sanity and gain !M! Block.",
+    "UPGRADE_DESCRIPTION": "At the start of each turn, lose !M! ${ModID}:Sanity. NL Whenever you lose *Sanity, gain !M! Block."
   },
   "${ModID}:EndOfDays": {
```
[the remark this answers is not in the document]
mikemayhemdev: So End of Days in my mind still plays indirectly with the Lunacy mechanic, in that if you go deep into Lunacy you can often find yourself with excess energy on turns at random depending when you end up getting low.  I like indirect plays like this for Rares more often than direct references for Packmaster as it is easier to draft when you aren't going deep into the mechanic, but a smart player can realize the indirect synergy and make use of it.

mikemayhemdev:
All comments from Modargo either resolved or commented on.

src/main/java/thePackmaster/powers/cthulhupack/SanityPower.java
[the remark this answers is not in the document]
modargo: The power icon makes sense as a reason -- minor and only aesthetic, but I get it and it's fine with me.

src/main/resources/anniv5Resources/localization/eng/dimensiongatepack3/Cardstrings.json
```diff
@@ -16,8 +16,8 @@
   },
   "${ModID}:Ouroboros": {
     "NAME": "Ouroboros",
-    "DESCRIPTION": "At the end of each turn, if you fully blocked any source of damage, gain 1 Strength.",
-    "UPGRADE_DESCRIPTION": "Innate. NL At the end of each turn, if you fully blocked any source of damage, gain 1 Strength."
+    "DESCRIPTION": "The first time you fully block any source of damage each turn, gain 1 Strength.",
+    "UPGRADE_DESCRIPTION": "Innate. NL The first time you fully block any source of damage each turn, gain 1 Strength."
```
modargo, line 20: Since this was changed from "blocked" to "block" it will also need the zero-width space treatment.
```suggestion
    "UPGRADE_DESCRIPTION": "Innate. NL The first time you fully block any source of damage each turn, gain 1 Strength."
```

src/main/resources/anniv5Resources/localization/eng/dimensiongatepack3/Cardstrings.json
```diff
@@ -16,8 +16,8 @@
   },
   "${ModID}:Ouroboros": {
     "NAME": "Ouroboros",
-    "DESCRIPTION": "At the end of each turn, if you fully blocked any source of damage, gain 1 Strength.",
-    "UPGRADE_DESCRIPTION": "Innate. NL At the end of each turn, if you fully blocked any source of damage, gain 1 Strength."
+    "DESCRIPTION": "The first time you fully block any source of damage each turn, gain 1 Strength.",
```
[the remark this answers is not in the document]
modargo: I'll just go ahead and make this edit through Github since it's very small.

src/main/resources/anniv5Resources/localization/eng/dimensiongatepack3/Cardstrings.json
```diff
@@ -8,28 +8,28 @@
...
   "${ModID}:Minniegun": {
     "NAME": "Minniegun",
     "DESCRIPTION": "Deal !D! damage !M! times. NL Put a *Jam in your discard pile."
   },
   "${ModID}:Ouroboros": {
     "NAME": "Ouroboros",
-    "DESCRIPTION": "At the end of each turn, if you fully blocked any source of damage, gain 1 Strength.",
-    "UPGRADE_DESCRIPTION": "Innate. NL At the end of each turn, if you fully blocked any source of damage, gain 1 Strength."
+    "DESCRIPTION": "The first time you fully block\u200b any source of damage each turn, gain 1 Strength.",
+    "UPGRADE_DESCRIPTION": "Innate. NL The first time you fully block any source of damage each turn, gain 1 Strength."
```
modargo, line 20:
```suggestion
    "UPGRADE_DESCRIPTION": "Innate. NL The first time you fully block\u200b any source of damage each turn, gain 1 Strength."
```

src/main/java/thePackmaster/cards/dimensiongatepack3/SelfWound.java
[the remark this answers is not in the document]
modargo: Can't hurt to see how it works in practice! Quite possibly it will turn out to be better than my initial impression of it.

src/main/resources/anniv5Resources/localization/eng/cthulhupack/Cardstrings.json
```diff
@@ -1,33 +1,33 @@
...
     "NAME": "Gaze the Void",
-    "DESCRIPTION": "Deal !D! damage !M! times. NL Enter ${ModID}:Nightmare."
+    "DESCRIPTION": "Deal !D! damage to a random enemy. NL Repeat for each *Lunacy played this combat.",
+    "EXTENDED_DESCRIPTION": [" NL (Hits ", " times.)"]
   },
   "${ModID}:AllSeeing": {
-    "NAME": "All Seeing",
-    "DESCRIPTION": "At the end of each turn, Scry !M! and take 3 damage."
+    "NAME": "Eldritch Pact",
+    "DESCRIPTION": "At the start of each turn, lose !M! ${ModID}:Sanity. NL Whenever you lose *Sanity, gain !M! Block."
```
modargo, line 30: A minor change to use `${ModID}:Sanity` instead of `*Sanity`. I'll just make it through Github.
```suggestion
    "DESCRIPTION": "At the start of each turn, lose !M! ${ModID}:Sanity. NL Whenever you lose ${ModID}:Sanity, gain !M! Block."
```

src/main/resources/anniv5Resources/localization/eng/cthulhupack/Cardstrings.json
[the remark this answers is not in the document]
modargo: It seems a bit too indirect for my tastes, but I don't feel too strongly about it.

src/main/resources/anniv5Resources/localization/eng/dimensiongatepack3/Cardstrings.json
```diff
@@ -8,28 +8,28 @@
   },
   "${ModID}:Inferno": {
     "NAME": "Inferno",
-    "DESCRIPTION": "Deal !D! damage and apply 10 ${ModID}:Ignite to ALL enemies. NL Gain !M! ${ModID}:Ignite."
+    "DESCRIPTION": "Deal !D! damage and apply !M! ${ModID}:Ignite to ALL enemies. NL Add 2 *Burns to your hand."
   },
   "${ModID}:Minniegun": {
     "NAME": "Minniegun",
     "DESCRIPTION": "Deal !D! damage !M! times. NL Put a *Jam in your discard pile."
```
modargo, line 15: Minniegun puts the Jam in your draw pile now, right? Looks like the text wasn't updated.
```suggestion
    "DESCRIPTION": "Deal !D! damage !M! times. NL Shuffle a *Jam into your draw pile."
```

src/main/resources/anniv5Resources/localization/eng/dimensiongatepack3/Cardstrings.json
```diff
@@ -8,28 +8,28 @@
...
   },
   "${ModID}:Ouroboros": {
     "NAME": "Ouroboros",
-    "DESCRIPTION": "At the end of each turn, if you fully blocked any source of damage, gain 1 Strength.",
-    "UPGRADE_DESCRIPTION": "Innate. NL At the end of each turn, if you fully blocked any source of damage, gain 1 Strength."
+    "DESCRIPTION": "The first time you fully block\u200b any source of damage each turn, gain 1 Strength.",
```
modargo, line 19: Going to make a set of edits to fix the block keyword thing in a slightly different way. The original way I suggested was behaving inconsistently for reason I don't fully understand.
```suggestion
    "DESCRIPTION": "The first time you fully block\u00a0any source of damage each turn, gain 1 Strength.",
```

src/main/resources/anniv5Resources/localization/eng/dimensiongatepack3/Cardstrings.json
```diff
@@ -8,28 +8,28 @@
...
+    "UPGRADE_DESCRIPTION": "Innate. NL The first time you fully block\u200b any source of damage each turn, gain 1 Strength."
```
modargo, line 20:
```suggestion
    "UPGRADE_DESCRIPTION": "Innate. NL The first time you fully block\u00a0any source of damage each turn, gain 1 Strength."
```

src/main/resources/anniv5Resources/localization/eng/dimensiongatepack3/Cardstrings.json
```diff
@@ -8,28 +8,28 @@
...
+    "DESCRIPTION": "Gain !B! Block. NL Add a *Jam to your hand."
   },
   "${ModID}:Jam": {
     "NAME": "Jam",
     "DESCRIPTION": "Take 3 damage. NL Draw a card. NL Exhaust."
   },
   "${ModID}:MantisStrike": {
     "NAME": "Mantis Strike",
-    "DESCRIPTION": "Take 3 damage. NL Deal !${ModID}:sd! damage. NL Deal !D! damage to a random enemy, twice."
+    "DESCRIPTION": "Deal !D! damage to a random enemy. NL Repeat whenever you fully block​\u200b a source of damage this turn."
```
modargo, line 32:
```suggestion
    "DESCRIPTION": "Deal !D! damage to a random enemy. NL Repeat whenever you fully block\u00a0a source of damage this turn."
```

src/main/java/thePackmaster/powers/dimensiongate3pack/MantisStrikePower.java
```diff
@@ -0,0 +1,51 @@
...
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class MantisStrikePower extends AbstractPackmasterPower {
+    public static final String POWER_ID = makeID("MantisStrikePower");
+    public static final String NAME = CardCrawlGame.languagePack.getPowerStrings(POWER_ID).NAME;
+    public static final String[] DESCRIPTIONS = CardCrawlGame.languagePack.getPowerStrings(POWER_ID).DESCRIPTIONS;
+
+    public MantisStrikePower(AbstractCreature owner, int damageAmount) {
+        super(POWER_ID, NAME, PowerType.BUFF, true, owner, 1);
```
modargo, line 22: It looks like `1` is hardcoded as the amount being passed to the superclass constructor here, which I think is causing a bug where Mantris Strike's buff does the wrong damage. Also, this shouldn't be marked as a turn-based power.
```suggestion
        super(POWER_ID, NAME, PowerType.BUFF, false, owner, damageAmount);
```

Once you make this change you can delete the `amount = damageAmount` and `updateDescription()` lines below, the superclass constructor takes care of all that.

src/main/resources/anniv5Resources/localization/eng/cthulhupack/Powerstrings.json
```diff
@@ -3,22 +3,28 @@
...
+      " #ySanity."
     ]
   },
-  "${ModID}:AllSeeingPower": {
-    "NAME": "All Seeing",
+  "${ModID}:AllSeeingPowerTriggerOnCard": {
+    "NAME": "Eldritch Pact+",
+    "DESCRIPTIONS": [
+      "Whenever you lose #ySanity, gain #b",
+      " Block."
```
erasels, line 20:
```suggestion
      " #yBlock."
```

src/main/resources/anniv5Resources/localization/eng/cthulhupack/Powerstrings.json
```diff
@@ -3,22 +3,28 @@
     "NAME": "Nameless Mist",
     "DESCRIPTIONS": [
       "Whenever you play a card this turn, gain #b",
-      " #yBlock."
+      " Temporary HP."
```
erasels, line 6:
```suggestion
      " #yTemporary #yHP."
```

src/main/resources/anniv5Resources/localization/eng/cthulhupack/Powerstrings.json
```diff
@@ -3,22 +3,28 @@
...
+    ]
+  },
+  "${ModID}:AllSeeingPowerTriggerOnTurn": {
+    "NAME": "Eldritch Pact",
     "DESCRIPTIONS": [
-      "At the end of each turn, Scry #b",
-      " and take ",
-      " damage."
+      "At the start of your turn, gain #b",
+      " Block."
```
erasels, line 27:
```suggestion
      " #yBlock."
```

PR #575, opened 2023-04-12
occult gem is occult :)

Alchyr:


erasels:
Why? You need to adjust the strings as well

PR #584, opened 2023-04-14
Graveyard pack: change Grey Bargain to not shuffle your draw pile

modargo:


erasels:
What's the reason for this change?

modargo:
> What's the reason for this change?

It might not be obvious but this is unintended behavior that is effectively a bug -- something the card shouldn't have been doing that results in unexpected outcomes from the perspective of a player who knows how the game systems work.

Base game precedent (and near-universal practice for modded cards too) is that shuffling your draw pile is a specific operation that only happens when its directly said on the card (beyond the normal shuffle when your draw pile is empty). See Reboot and Deep Breath for base game examples. Effects that shuffle something into your draw pile do _not_ shuffle your draw pile -- they simply put the shuffled in card in a random spot. Bunch of examples of that; Wild Strike, Reckless Charge, and Chrysaslis are three that come to mind immediately.

Grey Bargain is that type of effect -- it might be taking the card from your exhaust pile, but functionally it should work like any other card that shuffles cards into your draw pile. The difference matters -- if you've put cards in specific positions in your draw pile (e.g. with Sideboard) playing Grey Bargain shouldn't shuffle them around, and playing Grey Bargain shouldn't trigger Sundial/Abacus either. (I discovered this issue because I ran into the incorrect interaction with Sideboard.)

Looking at this code, I assume the card worked this way simply because of lazy programming rather than any deliberate choice. Whoever wrote this didn't realize that shuffling your library mattered and probably didn't know how to pass a callback to the action, so they just threw in a shuffle after the move cards action (since that action always puts the moved cards on top). But that's not how a card like this should work -- only the cards being moved should be affected, not your entire draw pile. The shuffle operation isn't mentioned by the card text either, it's just something that is hidden there, waiting to mess things up in the situations where it matters.

erasels:
Gotcha, thanks

PR #606, opened 2023-04-22
bugfixes/adjustments

Alchyr:
slash draw didn't use magic number

another moment would count both skills played *and* skills it had exhausted, which resulted in each skill counting for 2 instead of 1

lunacy's upgrade was missing text of "lose 1 sanity", when it only lost the hp loss effect

recode black market dealer event with thing I added to basemod
Also adjusted so it chooses 2 of 3 to be available instead of just 1 because it feels really pointless to just have a screen with 1 option

Make `AbstractDungeon.getRandomCard` return a card from all packmaster cards if pools are not yet initialized (fixes crash with insanity modifier for custom runs)

Add `None` option to custom draft. If someone enables only one pack and crashes it's their own fault :)
(needs chinese translation for UI option)


modargo:
I'm not sure we want the "none" option for custom draft. Having functionality that causes crashes when used in a natural way (e.g. turning off all but one or two packs like people have wanted to do) isn't good. Letting people do that without any warning implies that its a first-class feature that we fully support, just like everything else. But that's not true because of the many unsolvable crashes/freezes it can cause, but people will report crashes anyway.

I think the status quo of not having this option at all is a good place to be. It is okay for us to simply not address the feedback from people saying that they want to play with only one or two packs enabled. So my vote is to remove this from the PR.

However, if we really think that adding this option is a positive, we should at least put it behind a config option in the mod config menu with a label like "Allow less than 7 packs in custom draft (may cause crashes)", defaulting to off. That will appropriately convey that this is something we do not fully support and limit the use of a feature we don't really want to support. But I think it's simpler and better to just not include this at all.

erasels:
Looking at the code, there is not extra logic to disable the filtering so you can only disable 1 pack with the none option. 
I think I did the math for that at some point and it's not possible to run into a crash with that, so I'm fine with merging.

erasels:
I'll fix it myself

PR #610, opened 2023-04-24
Custom draft: warn that the option for selecting None multiple times may cause crashes/freezes

modargo:
We warned people about this in the Steam comment about the update. I'd like to have this warning in-game as well.

Leaving this for you to merge in case you want to discuss. My principle is that it's important to label things that may cause crashes and/or aren't fully supported, like this option. It helps set expectations.


erasels:
Yeah good point, i forgot adding that.

PR #619, opened 2023-04-29
Aggression and Shaman pack balance tweaks

modargo:
As usual with balance changes, leaving this for you to merge (or engage with if you want to). My thoughts are below. Thanks for sharing the pick data, it's a good way to identify problem areas!
* Aggression pack: increase Dark Lance damage and upgrade damage by 1
* Shaman pack: change Ice Wall to add a Dazed instead of a Void and reduce block and upgrade block by 1

This comes form taking a quick look over the pick data for my packs. Almost everything was in good shape, with most cards in the healthy 15% to 50% range. There were a few rares above that range and a few niche cards below it, but nothing that seemed too worrisome. Its fine for some strong rares to have high pick rates and it's fine for some cards to be more niche picks as long as it fits the context of the pack and character.

I focused my changes here on two cards that I felt were in bad shape:
* Dark Lance had a pick rate around 5%, similar to Flick. I set the numbers for this card with the idea that you'd trigger the bonus condition a reasonable amount of the time, but it turns out to be harder to do that than I anticipated. Stances aren't as common of a mechanic and you need to line up a stance change card and this and still have energy left over. I still like the card design, so I'm just adjusting the damage up slightly.
* Ice Wall had a pick rate around 2%, among the least picked cards. The Void is just too harsh of a drawback. In addition to making the card awkward in practice, it probably deterred people from picking the card even when it would be a decent fit for a particular deck. Given that, rather than just buffing the numbers, I've changed the card to add a Dazed instead of a Void. That's still a meaningful downside, but it opens up a lot more flexibility in how you handle it. I adjusted the numbers down to compensate, though since the previous version of the card wasn't very good, I only made a slight reduction (2 more block than Dodge and Roll in return for losing a draw next turn seems reasonable to me).

Both of these are cards that were already on my radar as too weak. Seeing the pick data validated that.

### DIFF
--- a/src/main/java/thePackmaster/cards/aggressionpack/DarkLance.java
+++ b/src/main/java/thePackmaster/cards/aggressionpack/DarkLance.java
@@ -16,8 +16,8 @@ import thePackmaster.patches.aggressionpack.DarkLancePatch;
 public class DarkLance extends AbstractAggressionCard {
     public static final String ID = SpireAnniversary5Mod.makeID("DarkLance");
     private static final int COST = 1;
-    private static final int DAMAGE = 9;
-    private static final int UPGRADE_DAMAGE = 2;
+    private static final int DAMAGE = 10;
+    private static final int UPGRADE_DAMAGE = 3;
     private static final int DRAW = 2;
     private static final int UPGRADE_DRAW = 1;
 
--- a/src/main/java/thePackmaster/cards/shamanpack/IceWall.java
+++ b/src/main/java/thePackmaster/cards/shamanpack/IceWall.java
@@ -12,7 +12,7 @@ import thePackmaster.SpireAnniversary5Mod;
 public class IceWall extends AbstractShamanCard {
     public static final String ID = SpireAnniversary5Mod.makeID("IceWall");
     private static final int COST = 1;
-    private static final int BLOCK = 6;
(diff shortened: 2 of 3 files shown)


erasels:
I'm fine with merging this, but are you really sure you want to leave Ice Wall in this state? Personally, I'd pick this even less than the original version. Dodge and Roll is already a card I would only pick if I have dex which is not that easy to comeb by in PM either.

I'm not exactly sure why I feel this way, but somehow taking a status for a defend now and next turn just really doesn't feel worth it to me. I'd probably increase the base block to 7

modargo:
> I'm fine with merging this, but are you really sure you want to leave Ice Wall in this state? Personally, I'd pick this even less than the original version. Dodge and Roll is already a card I would only pick if I have dex which is not that easy to comeb by in PM either.
> 
> I'm not exactly sure why I feel this way, but somehow taking a status for a defend now and next turn just really doesn't feel worth it to me. I'd probably increase the base block to 7

Let's try things out at 7 block, then. Doesn't hurt to make this a bit more attractive to pick. I'll keep the upgrade as only +2 though, don't want to make it too efficient now that it has less of a downside. 

I'll merge after making this change.

erasels:
Yeah, totally on board with the 2 block upgrade, looks good to me

PR #636, opened 2023-06-23
Nerfed Catharsis and reworded Burning Passion

Laugic:


src/main/java/thePackmaster/powers/rimworldpack/CatharsisPower.java
```diff
@@ -56,6 +60,10 @@ public void updateDescription() {
         else
             description = DESCRIPTIONS[1]+ amount + DESCRIPTIONS[2];
     }
+    @Override
+    public void atStartOfTurn() {
```
erasels, line 64:
```suggestion
    public void atEndOfTurn(boolean isPlayer) {
    if (isPlayer)
```

src/main/java/thePackmaster/powers/rimworldpack/CatharsisPower.java
```diff
@@ -56,6 +60,10 @@ public void updateDescription() {
...
+        addToTop(new RemoveSpecificPowerAction(owner, owner, this));
```
erasels, line 65:
```suggestion
        addToBot(new RemoveSpecificPowerAction(owner, owner, this));
```

erasels: This is the way Burst, etc do it. Might want to follow their example so its consistent

erasels (changes requested):

erasels:
I'll do it myself

PR #637, opened 2023-06-23
Various packs: fix Destiny Draw, Mimicry, More Explosions, Favor, and Retromation to work when played on other characters (e.g. if acquired through Prismatic Shard)

modargo:
After making the `receivePreStartGame` change from the previous PR, I looked through the Packmaster codebase for uses of `currentPoolPacks`. I found a few that I could tell would break if you played them on another character, where `currentPoolPacks` would be empty.

Very niche but figured I'd fix them... especially since a bunch of them could use the `Wiz.getCardsMatchingPredicate` function that does almost everything for you. That means this also simplified the code a lot. While I was at it I also fixed Destiny Draw -- the code changes weren't related, but it was another card I knew didn't work properly on other characters.

Leaving this one up for you to take a look over in case you're curious. At the very least, you might appreciate the code simplification.

### DIFF
--- a/src/main/java/thePackmaster/cards/coresetpack/DestinyDraw.java
+++ b/src/main/java/thePackmaster/cards/coresetpack/DestinyDraw.java
@@ -4,6 +4,7 @@ import com.megacrit.cardcrawl.cards.AbstractCard;
 import com.megacrit.cardcrawl.characters.AbstractPlayer;
 import com.megacrit.cardcrawl.monsters.AbstractMonster;
 import thePackmaster.SpireAnniversary5Mod;
+import thePackmaster.ThePackmaster;
 import thePackmaster.actions.FlexibleDiscoveryAction;
 import thePackmaster.cards.AbstractPackmasterCard;
 import thePackmaster.packs.AbstractCardPack;
@@ -22,7 +23,7 @@ public class DestinyDraw extends AbstractPackmasterCard {
     }
 
     public void use(AbstractPlayer p, AbstractMonster m) {
-        ArrayList<AbstractCardPack> choices = SpireAnniversary5Mod.getRandomPacks(true, 3);
+        ArrayList<AbstractCardPack> choices = SpireAnniversary5Mod.getRandomPacks(p.chosenClass == ThePackmaster.Enums.THE_PACKMASTER, 3);
 
         ArrayList<AbstractCard> packCards = new ArrayList<>();
 
--- a/src/main/java/thePackmaster/cards/creativitypack/Mimicry.java
+++ b/src/main/java/thePackmaster/cards/creativitypack/Mimicry.java
@@ -7,8 +7,8 @@ import com.megacrit.cardcrawl.characters.AbstractPlayer;
 import com.megacrit.cardcrawl.monsters.AbstractMonster;
 import thePackmaster.SpireAnniversary5Mod;
 import thePackmaster.actions.FlexibleDiscoveryAction;
-import thePackmaster.packs.AbstractCardPack;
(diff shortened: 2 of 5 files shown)


src/main/java/thePackmaster/cards/coresetpack/DestinyDraw.java
```diff
@@ -22,7 +23,7 @@ public DestinyDraw() {
     }
 
     public void use(AbstractPlayer p, AbstractMonster m) {
-        ArrayList<AbstractCardPack> choices = SpireAnniversary5Mod.getRandomPacks(true, 3);
+        ArrayList<AbstractCardPack> choices = SpireAnniversary5Mod.getRandomPacks(p.chosenClass == ThePackmaster.Enums.THE_PACKMASTER, 3);
```
modargo, line 26: `true` here means "pick from the pool of packs for my current run", `false` means "pick from all packs".

src/main/java/thePackmaster/cards/creativitypack/Mimicry.java
```diff
@@ -33,13 +33,7 @@ public void use(AbstractPlayer abstractPlayer, AbstractMonster abstractMonster)
...
             @Override
             public void update() {
-                ArrayList<AbstractCard> list = new ArrayList<>();
-                for (AbstractCardPack pack : SpireAnniversary5Mod.currentPoolPacks) {
-                    if (!CreativityPack.ID.equals(pack.packID)) {
-                        list.addAll(pack.cards);
-                    }
-                }
-                list.removeIf(c -> c.rarity != CardRarity.COMMON);
+                ArrayList<AbstractCard> list = Wiz.getCardsMatchingPredicate(c -> c.rarity == CardRarity.COMMON && !CreativityPack.ID.equals(SpireAnniversary5Mod.cardParentMap.getOrDefault(c.cardID, null)) && !c.hasTag(CardTags.HEALING));
```
modargo, line 36: This code was didn't have the check to exclude healing tagged cards previously, which nobody noticed because there probably aren't any common healing cards.

src/main/java/thePackmaster/powers/contentcreatorpack/RetromationPower.java
```diff
@@ -26,38 +27,30 @@ public class RetromationPower extends AbstractPackmasterPower {
...
 
-        } else {
-            for (AbstractCardPack p : allPacks) {
-                if (!SpireAnniversary5Mod.currentPoolPacks.contains(p) && !(p instanceof GemsPack)) {
-                    potentialPacks.add(p);
-                }
-            }
-        }
+    private List<AbstractCard> getPotentialCards() {
+        return CardLibrary.getAllCards().stream()
```
modargo, line 38: While Mimicry, More Explosions, and Favor all create cards from your current card pool, Retromation makes cards from outside your current card pool. Specifically it says "At the start of your turn, gain a card from a Pack not available in your draft."

If you're not playing Packmaster, there are no packs in your draft, so having this make cards from all Packmaster cards seems the most sensible. The code for that isn't as easy as for the other cards, but it's still a bit simpler than it was before.

src/main/java/thePackmaster/cards/goddessofexplosionspack/MoreExplosions.java
```diff
@@ -33,24 +29,10 @@ public MoreExplosions() {
 
     @Override
     public void use(AbstractPlayer p, AbstractMonster m) {
-        // I'll be moderately mad if there's an automated way of doing this.
```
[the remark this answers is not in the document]
erasels: lol

PR #638, opened 2023-06-26
Add easier method for creating pack summary

erasels:
Pack Summary is a nice feature, but it was pointed out it's a bit hard to use from the outside.

Since we'll likely have an expansion packs mod soon, I've added an easier way of adding your own pack summary while keeping the old style for compatibility sake. Although, I do think it's a bit cumbersome and might be worth removing the text file at some point. It's noticeable in the amount of packs that were PR'ed without a summary or tags. The error logging is a nice touch, but it's easy to miss since there are so many logging calls that early into the intialization.

### DIFF
--- a/src/main/java/thePackmaster/packs/MarisaPack.java
+++ b/src/main/java/thePackmaster/packs/MarisaPack.java
@@ -4,6 +4,7 @@ import com.megacrit.cardcrawl.core.CardCrawlGame;
 import com.megacrit.cardcrawl.localization.UIStrings;
 import thePackmaster.SpireAnniversary5Mod;
 import thePackmaster.cards.marisapack.*;
+import thePackmaster.summaries.PackSummary;
 
 import java.util.ArrayList;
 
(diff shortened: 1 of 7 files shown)


modargo:
Makes sense given that we want to split out some packs. FYI I set this up with all the data in one file so that I could add and manage all the ratings and tags in bulk, without needing to touch every pack file. It was much faster and easier that way, especially since I had all the ratings and tags in a spreadsheet.

src/main/java/thePackmaster/packs/MarisaPack.java
```diff
@@ -35,4 +36,12 @@ public ArrayList<String> getCards() {
         cards.add(EarthlightRay.ID);
         return cards;
     }
+
+    @Override
+    public PackSummary getPackSummary() {
+        PackSummary pSum = new PackSummary(4, 1, 3, 4,3);
+        pSum.setTags("Exhaust", "Attacks");
```
modargo, line 43: Since there's a specific list of supported tags, it might be easier for people if we make constants for each one. It wasn't needed with the previous way of defining this since it just needed to have the right entry in ui.json. 

There's already a constant for None (`PackSummaryReader.NONE_TAG`) because of how that was treated specially, so just need to add constants like that for the rest:
```
"Strength"
"Exhaust"
"Orbs"
"Stances"
"Discard"
"Debuffs"
"Attacks"
"Tokens"
"Powers"
```

We could also let people just put any string for a tag. The reason I set it up to only allow a specific set of tags is because the pack summary is most useful if it highlights things that are shared between packs. Things specific to one pack can be noted in the pack description.

modargo (approved):

OceanUwU:
#639 does this in a different way

erasels:
Due to #639 being merged and being a better way to solve it, this change is not obsolete

PR #639, opened 2023-06-27
Make it easier for other mods to register packs + change the way pack summaries are set

OceanUwU:
Pack summaries are now declared in the constructor, with something like this:
```java
super(ID, NAME, DESC, AUTHOR, CREDITS, new AbstractCardPack.PackSummary(5, 2, 2, 4, 3, "Strength", "Stances", "Attacks"));
```
The list of tags can be left empty if None is the only tag the pack will have.

I've edited every pack constructor to set its summary in this way, and deleted the summary txt files.

---
Before these changes, it took 6 patches to add a pack from my character mod using cards only taken from its class cards, making sure those cards displayed the pack name on top when necessary and showing up in the packmaster section in the collection. With these changes, you no longer need any patches, and it looks something like this:

In the constructor of the main mod file, after subscribing to BaseMod:
```java
if (Loader.isModLoaded("anniv5"))
    SpireAnniversary5Mod.subscribe(new PackLoader());
```

In `PackLoader.java`:
```java
public class PackLoader implements EditPacksSubscriber {
    @Override
    public void receiveEditPacks() {
        SpireAnniversary5Mod.allowCardClass(AbstractCrankyCard.class);
        SpireAnniversary5Mod.allowCardColor(Cranky.Enums.CLOCKWORK_BROWN_COLOR);
        new AutoAdd(CrankyMod.modID)
            .packageFilter("clockworkchar.packs")
            .any(AbstractCardPack.class, (info, pack) -> SpireAnniversary5Mod.declarePack(pack));
    }
}
```

In the pack files everything looks mostly the same as packs would look when made in this mod, but `makePreviewCard()` is overridden to get the correct resource path for the pack preview card, and `getHatPath()` (a new method) is overridden to get the correct resource path for the pack's hat.


erasels:
Looks good, thank you for your contribution

PR #641, opened 2023-06-27
Change PackSummary to use enums instead of strings

erasels:
The old system just took in strings and would just crash if the given tag string didn't exist, making this system a little stricter via the use of enums should be good at guiding people to tag correctly and gives an easy overview of the current tags.


modargo:
Makes sense. Now that the pack summary is defined in code instead of the text file, this is a better way to do things.

Would be good to let ocean know since this will require adjustment to the Packmaster integration code.

modargo (approved):

PR #642, opened 2023-06-28
Packmaster Metrics

erasels:
Adds expanded metrics for the packmaster.

Aside from the normal run metrics it adds the following PM specific datapoints:

- Currently running PM version
- Current packs
- which packs were filtered
- What pack choices they got at the start of the run and which of them were picked
- if custom draft was enabled and what the configuration was if so
- the picked hat 🤠 
- If the as of yet unreleased exapnsion packs was enabled


src/main/java/thePackmaster/patches/MetricsPatches.java
```diff
@@ -0,0 +1,110 @@
...
+                    //Pack choices at the beginning formatted as Pack1|*Pack2|Pack3,Pack1... * signifying the selected pack
+                    addDataMethod.invoke(metrics, "packChoices", String.join(",", packChoices));
+                    boolean draftEnabled = SpireAnniversary5Mod.getCustomDraftEnabled();
+                    addDataMethod.invoke(metrics, "customDraftEnabled", draftEnabled);
+                    if(draftEnabled)
+                        addDataMethod.invoke(metrics, "customDraftConfiguration", SpireAnniversary5Mod.modConfig.getString("PackmasterCustomDraftSelection"));
+
+                    addDataMethod.invoke(metrics, "pickedHat", SpireAnniversary5Mod.getLastPickedHatID());
+                    addDataMethod.invoke(metrics, "enabledExpansionPacks", Loader.isModLoaded("expansionPacks")); //TODO: Actually use the correct field once both branches are merged
+
```
modargo, line 88: Another data point that would be interesting: whether packs added by mods other than Packmaster and its expansions exist. In other words, how often do people actually run stuff like ocean's mod alongside Packmaster (whether or not they actually use the new packs).

We should be able to tell that by looking at `SpireAnniversary5Mod.unfilteredAllPacks` and checking the namespace that the packs come from. In fact we might be able to just use `WhatMod.findModID`:
```
String expansionModID = "expansionPacks"; // We should probably put this in the main mod file alongside modID, if it's going to be used in multiple places
List<String> packmasterModIDs = Arrays.asList(SpireAnniversary5Mod.modID, expansionModID);
String otherModPackIDs = SpireAnniversary5Mod.unfilteredAllPacks.stream()
		.filter(p -> !packmasterModIDs.contains(WhatMod.findModID(p.getClass())))
		.map(p -> p.packID)
		.collect(Collectors.joining(","));
SpireAnniversary5Mod.logger.info("Packs from other mods: " + otherModPackIDs);
```

I tried that out with ocean's mod running and got `clockworkchar:ToolsPack` as the output.

src/main/java/thePackmaster/screens/PackSetupScreen.java
```diff
@@ -101,6 +103,7 @@ public void open(int randomPacks, int packChoices) {
         rng = new Random(Settings.seed);
 
         packPool.clear();
+        MetricsPatches.packChoices.clear();
```
modargo, line 106: There's one other place that we should clear this list -- in `OpeningRunScreenPatch.SetTheThing`, in the if branch for all packs mode. That branch never opens up the pack selection screen, and as currently written it would leave the pack choices data the same as the previous packmaster run (e.g. if you did a normal run, then turned on all packs mode and did another run).

Now that I think about it, this might be another case where you could make one clear call in `SpireAnniversary5Mod.receivePreStartGame` instead. That works for `currentPoolPacks`, so it should work for this.

src/main/java/thePackmaster/patches/MetricsPatches.java
```diff
@@ -0,0 +1,110 @@
...
+    @SpirePatch(clz = Metrics.class, method = "run")
+    public static class RunPatch {
+        @SpirePostfixPatch
+        public static void patch(Metrics metrics) {
+            if (metrics.type == Metrics.MetricRequestType.UPLOAD_METRICS && AbstractDungeon.player.chosenClass == THE_PACKMASTER) {
+                try {
+                    Method addDataMethod = Metrics.class.getDeclaredMethod("addData", Object.class, Object.class);
+                    addDataMethod.setAccessible(true);
+                    addDataMethod.invoke(metrics, "pmversion", findTheModVersion());
+                    addDataMethod.invoke(metrics, "currentPacks", SpireAnniversary5Mod.currentPoolPacks.stream().map(p -> p.packID).collect(Collectors.joining(",")));
```
modargo, line 71: Do we want to do anything special to detect all packs mode? Right now it will just record the list of all 90 packs (or however many exist for the player).

It's easy enough to look at the data and work out when all packs mode was on, so I don't have a strong opinion here.

src/main/java/thePackmaster/patches/MetricsPatches.java
```diff
@@ -0,0 +1,110 @@
...
+
+                    //Pack choices at the beginning formatted as Pack1|*Pack2|Pack3,Pack1... * signifying the selected pack
+                    addDataMethod.invoke(metrics, "packChoices", String.join(",", packChoices));
+                    boolean draftEnabled = SpireAnniversary5Mod.getCustomDraftEnabled();
+                    addDataMethod.invoke(metrics, "customDraftEnabled", draftEnabled);
+                    if(draftEnabled)
+                        addDataMethod.invoke(metrics, "customDraftConfiguration", SpireAnniversary5Mod.modConfig.getString("PackmasterCustomDraftSelection"));
+
+                    addDataMethod.invoke(metrics, "pickedHat", SpireAnniversary5Mod.getLastPickedHatID());
+                    addDataMethod.invoke(metrics, "enabledExpansionPacks", Loader.isModLoaded("expansionPacks")); //TODO: Actually use the correct field once both branches are merged
```
modargo, line 87: Making the ID of the expansion mod a constant in the main mod file would be good. Maybe you've already done that on the branch where you're working on it, but still good to do it here too! Easier to change all references if the name changes during development that way.

src/main/java/thePackmaster/patches/MetricsPatches.java
modargo: Now that I started thinking about all packs mode, another thing we could log is the other config stuff. There are four things you can change from the config menu itself, and there's also the pack summary checkbox on the pack selection screen.
```
SpireAnniversary5Mod.allPacksMode
SpireAnniversary5Mod.oneFrameMode
SpireAnniversary5Mod.sharedContentMode
SpireAnniversary5Mod.allowMultiNone()
SpireAnniversary5Mod.showPackSummaries()
```

Some of these aren't very interesting, but there's some value in knowing how often they're changed. We'd have empirical data about how often people actually changed stuff from the mod settings panel, at least.

This is another thing where I don't have a strong opinion, just doing my best to highlight data that may be worth logging so that we don't unintentionally leave out something that would have been valuable.

erasels: Pack summaries might be worth logging, good point. Although, since it's toggleable on the pack selection screen as well, it won't really tell us about the mod settings.

The others just aren't very interesting or worth logging in my opinion. Multi none can be easily seen in the custom run config log since you would see multiple nones there. I don#t even know what shared content mode is tbh, probably shares relics or something? Wonder when that came in

modargo: I agree, most of the settings aren't very interesting, and the interesting ones we can see the effects of by looking at other stuff we're already logging.

Some of the Packmaster relics were initially set up to be available for all characters. I believe MichaelMayhem liked the idea but since it's strongly disliked by many people we changed the relics to be Packmaster only, but added the shared content toggle so that you could opt in to them showing up on other characters. We probably had a brief discussion about it at some point during the initial development phase.

modargo (approved):
It'll be cool to have usage data for Packmaster. I can take a look over any changes that come out of my comments but I trust things will end up in a reasonable place.

src/main/java/thePackmaster/patches/MetricsPatches.java
erasels: I think a simpler solution would be to just check for anniv5: modid prefix for the pack.

PR #645, opened 2023-07-01
Fueled Pack

bkerns724:
comment


modargo:
Wanted to make sure you saw what Gk said in Discord about the new expansion pack repo -- new pack PRs should go there instead of here. You should be able to clone https://github.com/erasels/Packmaster-Expansion-Packs and copy all the code you have over to it (the structure should be the same as before, just in the new repo). Then you can make a PR there.

PR #651, opened 2023-07-24
Fix to IceShatterPower

ActuallyEnbeon:
Hi Gk (and Modargo if you are also here)
So I wanted to make a start on fixing this quickly - can't actually debug at all and I don't have the dependencies here with me but this might be a start?
Will continue work when I can, assuming you haven't got to it first that is!


ActuallyEnbeon:
By the way this code might be totally faulty in loads of ways but I will check when I can

ActuallyEnbeon:
Ok I'm home now and have tested this (it actually works now for realsies)

PR #656, opened 2023-08-13
Update to grand legacy and weapons packs

JohnnyBazooka89:
2 changes:

1. Small buff to 'Captain's Cabin' to match the current value of block in the Blackbeard mod. This card is supposed to have similar value to base game card 'Dash' and with this change it should be more with line with that card.
2. Small change to 'Sword' art, basically the swords were moved a few pixels down. Not very noticable, but again it's a change to match the current art in the Blackbeard mod.


erasels:
The retired modder special :)

PR #675, opened 2023-11-24
Balance changes and bug fixes

modargo:
Leaving this for you to take a quick look over and merge -- we discussed all the balance changes, but there are enough changes that it makes sense.

### Balance changes
Gems pack: change Citrine to rare
Rip pack: reduce Fragile Shrug block by 1
Rip pack: increase Inspiration cost by 1
Utility pack: increase Greater Hex cost by 1
Go With The Flow pack: change Flow to draw next turn equal to half the number of discarded cards (instead of equal to the number of discarded cards)
Go With The Flow pack: increase Hydraulics damage by 2
Go With The Flow pack: increase Aqueducts damage by 1
Creativity pack: reduce Makeshift Shield block by 1
Creativity pack: change Souvenirs upgrade to upgrade the created card (instead of creating an additional card)
Creativity pack: change Striking Strike to cost 2 and to upgrade to cost 1

### Fixes
Creativity pack: make the animation for Mimicry more consistent with other cards that discover a new card
Monster Hunter pack: fix Hunting Instincts showing the wrong block amount when you hold the card and hover it over the playing area
Replicators pack: fix Iterative Design being incorrectly marked as turn-based
Utility pack: tweak text formatting for Lesser Hex and Greater Hex to have each effect on its own line
Legacy pack: add keyword highlighting for Slow

### DIFF
--- a/src/main/java/thePackmaster/cards/creativitypack/Mimicry.java
+++ b/src/main/java/thePackmaster/cards/creativitypack/Mimicry.java
@@ -47,7 +47,7 @@ public class Mimicry extends AbstractCreativityCard {
                     }
                     return o;
                 }).collect(Collectors.toCollection(ArrayList::new));
-                addToBot(new FlexibleDiscoveryAction(JediUtil.createCardsForDiscovery(tmpGrp), false));
+                addToTop(new FlexibleDiscoveryAction(JediUtil.createCardsForDiscovery(tmpGrp), false));
                 isDone = true;
             }
         });
--- a/src/main/java/thePackmaster/cards/monsterhunterpack/HuntingInstincts.java
+++ b/src/main/java/thePackmaster/cards/monsterhunterpack/HuntingInstincts.java
@@ -27,8 +27,8 @@ public class HuntingInstincts extends AbstractMonsterHunterCard {
     }
 
     @Override
-    public void applyPowers() {
-        super.applyPowers();
+    public void applyPowersToBlock() {
+        super.applyPowersToBlock();
 
         if(hasHuntTarget()) {
             block *= 2;
@@ -37,7 +37,7 @@ public class HuntingInstincts extends AbstractMonsterHunterCard {
     }
 
     public void use(AbstractPlayer p, AbstractMonster m) {
-        addToBot(new GainBlockAction(p, block * (hasHuntTarget()? 2 : 1)));
+        addToBot(new GainBlockAction(p, block));
     }
 
     public boolean hasHuntTarget() {
--- a/src/main/java/thePackmaster/cards/rippack/Inspiration.java
+++ b/src/main/java/thePackmaster/cards/rippack/Inspiration.java
@@ -22,7 +22,7 @@ public class Inspiration extends AbstractRipCard {
 
 
     public Inspiration() {
-        super(ID, 0, CardType.SKILL, CardRarity.UNCOMMON, CardTarget.NONE);
+        super(ID, 1, CardType.SKILL, CardRarity.UNCOMMON, CardTarget.NONE);
         baseDamage = damage = 12;
         baseMagicNumber = magicNumber = 12;
         exhaust = true;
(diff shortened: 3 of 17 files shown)


src/main/java/thePackmaster/cards/creativitypack/Mimicry.java
```diff
@@ -47,7 +47,7 @@ public void update() {
                     }
                     return o;
                 }).collect(Collectors.toCollection(ArrayList::new));
-                addToBot(new FlexibleDiscoveryAction(JediUtil.createCardsForDiscovery(tmpGrp), false));
+                addToTop(new FlexibleDiscoveryAction(JediUtil.createCardsForDiscovery(tmpGrp), false));
```
modargo, line 50: With this change, the card stays up in the background while you're doing the discovery action, which is the same way it works for Wish/Discovery/etc. Need to use addToTop to achieve that due to nested actions.

I'm so used to some aspects of how the base game does things that stuff like this is disconcerting to me if I see it a lot. That's part of how I notice small consistency issues like this.

src/main/java/thePackmaster/cards/monsterhunterpack/HuntingInstincts.java
```diff
@@ -27,8 +27,8 @@ public void triggerOnGlowCheck() {
     }
 
     @Override
-    public void applyPowers() {
-        super.applyPowers();
+    public void applyPowersToBlock() {
```
modargo, line 30: The issue with the previous code was that `applyPowers` was changing the block, but `calculateCardDamage` wasn't. Since `calculateCardDamage` overwrites damage/block values when a card either acquires a target or is played, this meant that the numbers showed up wrong when the card acquired a target (and it's why the use action needed a `hasHuntTarget` check).

Overriding both `applyPowers` and `calculateCardDamage` would also fix this, but the nice thing about `applyPowersToBlock` is that it gets called by both those methods. Convenient if you're doing something that only affects block.

src/main/java/thePackmaster/cards/rippack/Inspiration.java
```diff
@@ -22,17 +22,15 @@ public class Inspiration extends AbstractRipCard {
...
+        super(ID, 1, CardType.SKILL, CardRarity.UNCOMMON, CardTarget.NONE);
         exhaust = true;
         CardModifierManager.addModifier(this, new RippableModifier());
     }
 
     @Override
     public void applyPowers() {
         super.applyPowers();
-        int statusAndArtCardsInExhaust = AbstractDungeon.player.exhaustPile.group.stream().filter(card -> isArtCard(card) || card.type == CardType.STATUS).collect(Collectors.toList()).size();
+        int statusAndArtCardsInExhaust = (int) AbstractDungeon.player.exhaustPile.group.stream().filter(card -> isArtCard(card) || card.type == CardType.STATUS).count();
```
modargo, line 33: Couldn't resist some drive-by simplifcation.

src/main/java/thePackmaster/powers/replicatorspack/IterativeDesignPower.java
```diff
@@ -26,7 +26,7 @@ public class IterativeDesignPower extends AbstractPackmasterPower implements Clo
 
 
     public IterativeDesignPower(AbstractCreature owner, int amount) {
-        super(POWER_ID, NAME, AbstractPower.PowerType.BUFF, true, owner, amount);
+        super(POWER_ID, NAME, AbstractPower.PowerType.BUFF, false, owner, amount);
```
modargo, line 29: My crusade continues.

src/main/resources/anniv5Resources/localization/eng/utilitypack/Cardstrings.json
```diff
@@ -32,12 +32,12 @@
   },
   "${ModID}:LesserHex": {
     "NAME": "Lesser Hex",
-    "DESCRIPTION": "Apply !M! Weak and !M! Vulnerable. NL Enemy loses !M! Strength. NL Add a *Greater\u00a0Hex to your discard pile. NL Exhaust.",
+    "DESCRIPTION": "Apply !M! Weak. NL Apply !M! Vulnerable. NL Enemy loses !M! Strength. NL Add a *Greater\u00a0Hex to your discard pile. NL Exhaust.",
```
modargo, line 35: This change makes the card text look better and matches how Uppercut is written. AlexMdle pointed it out to me.

erasels:
Looks great

PR #695, opened 2024-02-03
Come at me twitch chat. Ghost Claw scaling -1

mikemayhemdev:
Ghost Claw scaling -1

### DIFF
--- a/src/main/java/thePackmaster/cards/clawpack/GhostClaw.java
+++ b/src/main/java/thePackmaster/cards/clawpack/GhostClaw.java
@@ -19,7 +19,7 @@ public class GhostClaw extends AbstractClawCard {
     public GhostClaw() {
         super(ID, 0, CardType.ATTACK, CardRarity.SPECIAL, CardTarget.ENEMY, CardColor.COLORLESS);
         baseDamage = 3;
-        baseMagicNumber = magicNumber = 2;
+        baseMagicNumber = magicNumber = 1;
         tags.add(CLAW);
         exhaust = true;
     }


erasels:
I'm not sure if I like this change. The claw pack is a very insular pack by its main design and a big meme so streamers will commit to it, yes, but it's not really good.
It has one of the highest deviations between win rate between A0 and A20 and its win rate isn't especially high.
Metrics available here:
https://docs.google.com/spreadsheets/d/1-B-jM7OhSt1ZzFwBzo5DTe3gyIMoyARG_w4SInLybeA/edit#gid=0

Making it even slower might just make this pack not even worth picking at all for serious runs

mikemayhemdev:
While this is possible, I would argue a psychological counterpoint - that because the pack has such a high winrate in a0 for such a low amount of skill effort, players would tend to pick it expecting to have the same level of breeze which is not applicable in a20. By making the pack at a0 not in the top 10 of winrates in exchange for almost no piloting skill requirement, it will make the pack less picked in a20 as an assumed 'free win'. It will take more skill to pilot and synergize with other packs in order to find true success, and thus its differential I believe could decrease just as likely as it could increase.

erasels:
You raise an interesting point, but I'm not sure the assumption that people expect a breeze on A20 with this pack is the only reason its winrate is low there. Slowly scaling just isn't an option due to the increase punishment the enemies dish out, so the pack will always struggle.

The pack has a high win rate because the way to play with it is very intuitive "take every card from this pack you see" and it pays dividends. This changes on A20, to find a combo that allows you to play a few of these cards over and over while shoring up your defenses.

I fear that nerfing one of the main ways of making this pack scale will result in quite a bit of negative player sentiment and make the pack less fun in general.

Maybe @modargo can weigh in on this, I generally double check bigger balance decisions with him since I'm a low asc enjoyer. Might be a bit before he answers since he's travelling right now though.

mikemayhemdev:
You might be right on it being negative player sentiment (that will always happen with any nerfs). The best way to scale now is through 'easy' scalars my opinion, particularly through Infinite Claws (which in reality is a much much better Infinite Blades as currently printed), and through Cloak and Claw (also much better than Cloak and Dagger by comparison). Ghost Claw 'for the luls' replaces Shivs, but they are just better than Shivs right now in all cases except where you drafted only one source and nothing else. By nerfing the token, it puts more pressure on drafting and utilizing the actual Claws which now have the clearly higher Sharpen values.

That all being said, I'd say the same thing if I were in your position most likely. It's a beloved pack because of stream meme, I just hate to see it also be a top 10 winrate pack that requires very little piloting skill.

modargo:
tl;dr: I would want to keep Claw pack unchanged since it's already a bit weak for high ascension play and I'm not as concerned about the other aspects of it. Deeper explanation below.

I think Claw pack being weaker on A20 is more about the pack not being very good at the challenges A20 creates. At low ascensions you can win with nothing but offense; I've seen people speedrunning glitchless Ironclad purely by getting an early Whirlwind (it's enough to take you through act 3 with a bit of help). That kind of thing works at A0, but doesn't work at all for A20 heart run due to the combination of extra HP, more incoming damage, and harder enemy abilities.

This same dynamic happens with some other offense-focused packs, like Aggression. There are other reasons a pack can fall off between A0 and A20, but having lots of offense and not much else is a major one. Packs like that let people win at A0 by taking almost nothing other than its cards, but that stops working when facing a more difficult challenge.

I do agree that Claw pack gets the attention because of how popular Claw is (and all the memes around it, Claw is law caw caw and so on). The way I see things, though, Claw pack is in a similar to position to other packs when you look at the data. Other packs are close to it in terms of the A0/A20 comparison, most are ahead of it on A20 win rate, and it's not an outlier when you view it as one of the packs in the high A0 win rate/big fall off cluster. 

As part of that cluster and having a below-average A20 win rate, I don't think the pack needs nerfs. If it had a high A20 win rate things would be different, but nerfing a pack that's already a bit weak at A20 is rough. For the record, I agree with the sentiment that it's too easy of a pack on low ascension (and so are a few other packs). It has a straightforward plan that's usually good enough. But I feel it's okay to have a few packs like that around, and when those packs are weak at high ascension (as Claw is) and have a fun play experience (as Claw does for many people), I think we're better off without nerfs. (I feel the same based on my own experience playing Claw pack at A20).

People who are playing casually will have their fun with Claw pack, and people who are a bit more serious will either play at a high enough ascension that it no longer dominates or turn the pack off. As a committed A20 player, I've turned several packs off myself (the ones that I see as far too weak for A20 play), so a committed low ascension player turning off Claw pack seems like a reasonable thing too. With how hard it is to get every pack balanced properly at both high and low ascension, letting the players who play many Packmaster runs decide which packs they're not okay with is a philosophy I've come to embrace. It lets us have packs that are fun but imperfect like this, making the most of their positive impact while letting people opt out as desired.

mikemayhemdev:
Alright, I contend that it may not be worth nerfing. But, one final shot at defense of it:

Modargo agrees that it is too easy of a pack on low ascension, which is where my personal play is and where my nerf comes from, that the Ghost Claws feel too free of scaling. The main argument against I would turn back around - if the pack is already not viable at a20 due to not scaling fast enough or not having sufficient defenses, then I don't think such a small change would make a significant impact. If the argument is "people who are a bit more serious will either play at a high enough ascension that it no longer dominates or turn the pack off", then I don't think its a20 viability is a strong defensible point. If it is a worthwhile improvement for a0, which I still contend is the case due to its ease of piloting, then I think it still may have merit. And I'd be happier having Claw pack not in the top 10 winrate for the area where most players play, especially if it's already bad enough that it's not a20 viable anyway.

But, the argument of "its beloved and players might react poorly" may trump any tuning merit anyway. As I say often unto others about my own projects, "it's not mine anymore, it's the player's". So, if you both feel strongly, so shall it remain. Feel free to kill the PR and just do the Reprogram change separately.

mikemayhemdev:
No slight at all, I'm a professional after all. 90% of my proposals don't make it in. Discussing the ones that don't still matter.

I should have clarified the 0-cost comment more, I did mean "claws and 0-cost attacks". Added synergy functionality on top that wouldn't affect the pack itself, just add more reasons why you might care about it in relation to other packs, instead of it for the most part shoving out every other attack available in the pool.

PR #709, opened 2024-06-14
Creativity pack: fix Accumulative Strike damage not increasing when cards are created

modargo:
The issue is that `OnCreateCardInterface` isn't a subscriber, but the main mod file tried to use it like one. Looking at StSLib, it only works on relics/powers/cards/monsters. https://github.com/kiooeht/StSLib/blob/master/src/main/java/com/evacipated/cardcrawl/mod/stslib/StSLib.java#L226


erasels:
Oops, that one's on me.

PR #711, opened 2024-06-23
Feature/marisa reworkv3

erasels:
Still requires art, this is just for code review.

This is a massive rework of the Marisa pack as dicussed in the Marisa PM Pack Discussion thread. I'm keeping in old cards to not affect current runs.

image


erasels:
With art:
image

src/main/java/thePackmaster/powers/marisapack/FreeAmplifyPower.java
```diff
@@ -0,0 +1,42 @@
...
+import basemod.interfaces.CloneablePowerInterface;
+import com.megacrit.cardcrawl.cards.AbstractCard;
+import com.megacrit.cardcrawl.core.CardCrawlGame;
+import com.megacrit.cardcrawl.powers.AbstractPower;
+import thePackmaster.powers.AbstractPackmasterPower;
+import thePackmaster.util.Wiz;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class FreeAmplifyPower extends AbstractPackmasterPower implements CloneablePowerInterface, AmplifyPowerHook {
```
modargo, line 12: While reviewing this it hit me that since this power doesn't expire at the end of the turn, you will always get value from Blazing Star as long as you have cards with Amplify in your deck. It just might be delayed until you draw your next Amplify card.

This also means that the more Amplify cards you have, the more Blazing Stars you want, and the more Blazing Stars you have, the more Amplify cards you want. A virtuous cycle. Card draw would be the third piece of the puzzle, so it might be a good thing that Marisa doesn't have much, means you have to look to other packs to really fuel this kind of play.

Makes me really interested to see how well this works in practice.

src/main/java/thePackmaster/cards/marisapack/MasterSpark.java
```diff
@@ -1,46 +1,82 @@
...
+            calculateCardDamage(null);
+        }
+        allDmg(AbstractGameAction.AttackEffect.NONE);
+        if (isAmplified(this)) {
+            baseDamage = tmpBaseDamage;
+            applyPowers();
+        }
+    }
+
+    // lots of jank here because using secondDamage will make the multiDamage array not upgrade
```
modargo, line 37: To be fair, this is similar to what the base game does for cards like Halt. Which I agree is a bit jank, but it's jank with base game precedent that reliably works, which is something I'm on board with.

src/main/java/thePackmaster/powers/marisapack/ChargeUpPower.java
```diff
@@ -18,8 +18,11 @@ public class ChargeUpPower extends AbstractPackmasterPower implements CloneableP
     public static final String[] DESCRIPTIONS = CardCrawlGame.languagePack.getPowerStrings(POWER_ID).DESCRIPTIONS;
     public static int THRESHOLD = 6, MODIFY_AMT = 2;
 
+    public static int chargeUpGainThisCombat = 0;
+
     public ChargeUpPower(int amount) {
         super(POWER_ID, NAME, PowerType.BUFF, false, Wiz.p(), amount);
+        chargeUpGainThisCombat += amount;
```
modargo, line 25: I recall having logic like this in the constructor causing issues with Together in Spire. Not the biggest of compatibility issues, but it might be better to have this logic in both `onInitialApplication` and `stackPower` instead.

modargo (approved):
Looks good. Only a few minor suggestions.

src/main/java/thePackmaster/patches/marisapack/AmplifyPatches.java
```diff
@@ -20,6 +20,8 @@
 
 public class AmplifyPatches {
     public static AbstractCard amplified = null;
+    public static boolean amplifiedThisTurn = false, amplifiedThisCombat = false;
```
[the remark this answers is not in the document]
erasels: Yeah, thought about that as well. Figured it'd just be nicer to have a boolean to check though.

src/main/java/thePackmaster/powers/marisapack/ChargeUpPower.java
erasels: That was my original approach, but I figured I'd keep it the same as Mantra. Guess I'll switch back

src/main/java/thePackmaster/powers/marisapack/FreeAmplifyPower.java
erasels: Yeah, I always saw this as a Flex + delayed energy gain. I was hoping this would be a good incentive to go for Amplify cards.

PR #712, opened 2024-06-24
Anomaly pack: fix Voracious Sigil putting a copy of the exhausted card in your discard pile

modargo:


erasels:
Wait, this was an issue? How has nobody ever mentioned this...

modargo:
> Wait, this was an issue? How has nobody ever mentioned this...

Surprising what kind of issues people can miss, isn't it? I noticed it after I started picking the card more (because of the Discord discussion thread for it a while back where I thought it was mediocre and other people said it was really good). Previously I'd almost never picked it.

My assumption is that the extra card being in your discard meant nobody ever realized it when when they'd just played Voracious Sigil, and then didn't notice when the extra card came up in their next shuffle. If you're not paying close attention the delay makes it easy to overlook. And paying close attention to what should happen is a big part of finding bugs, it's part of why I find so many.

PR #725, opened 2024-08-30
Pack filter menu: make button positioning consistent for different resolutions

modargo:


src/main/java/thePackmaster/patches/MainMenuUIPatch.java
```diff
@@ -71,7 +71,7 @@ public class MainMenuUIPatch {
     // We pass these values to the button class, which does its own multiplication by xScale/yScale
     // Since we want this to be offset by a constant amount from the top of the screen, we divide by yScale to cancel it out
     private static final float HATBUTTON_X = 610f;
-    private static final float HATBUTTON_Y = (Settings.HEIGHT - 80f) / Settings.yScale;
+    private static final float HATBUTTON_Y = (Settings.HEIGHT - 120f * Settings.yScale) / Settings.yScale;
```
erasels, line 74: Won't this cause issues because you multiply the value by settings scale and then dievide it by that as well?

modargo: The key is that the button class takes this and multiplies it by `Settings.yScale` (as noted by the comment a few lines above), so that cancels out the division. The button ends up with a y-position of `Settings.HEIGHT - 120f * Settings.yScale`, which is exactly what we want.

The button doing its own multiplication by xScale/yScale is what makes this weird. I sort of wish it didn't do that, but that's how the BaseMod classes were built.

erasels: Ahh right, I remember that weirdness. BaseMod UI is a bit of a pain to deal with, haha.

Issue #731, opened 2024-11-02
Echo not working as expected


MystWraith:
_This might be intentional for balance purposes as you might see in a moment._
_Thought I'd also put it here, I don't know how often the steam discussion would be checked._

# Packs
- Energy & Echo
- A pack with any kind of Modify effect.

# Steps
1. Enhance/Modify a card with echo.
2. Play that card.
3. Notice that the echoed version lacks modifications.

# Explanation
**Echo** is described as "Add a copy of this card to your hand with **exhaust** and **ethereal**." I believe this would extend to most* modifications as it is now "part of the card" because the card "gains" that effect or text (as per the definition--or rather game given explanation--of how modify/enhance function).

I would accept that this could be a balancing choice (as putting something like "**channel** 1 **lightning**" on an **echo** card is quite strong), but if that were the case, could the tooltip for **echo** be changed to something like "Put an unmodified copy of this card into your hand, it gains **exhaust** and **ethereal**"?

\* **Stitching** would not apply as it attaches to the EXACT card, similar to how genetic algorithm copies would not increase the value of the original.

erasels:
Hey, we do check the steam discussions pretty regularly, and I have seen your message there, however I don't mind people opening issues. Personally, I like issues more because they're much easier to keep track of.

First of all, thank you for your detailed issue! I believe originally Echo used AbstractCard:makeStatEquivalentCopy to make a copy of the card which would work as you expect it, copying card modifiers and other such modifications to the card. However, due to balance reasons (specifically easy infinite) all cards that made copies of others were switched to not do that. I don't quite remember what exactly the infinite was here, as exhaust and ethereal should take care of it, but there was something.

As for renaming the tooltip, we decided not to go that route as we call this effect copying in all other instances (take Eternal Strike for example) mostly due to text space issues. It's a bit of a rough edge that will take people by surprise once, and we're not happy, but this was the solution we landed on.

PR #732, opened 2024-11-04
Energy and Echo pack: clarify Echo keyword to say that it creates an unmodified copy

modargo:


erasels:
I guess that'll solve #731 then. Personally, I don't think it was needed since we had precedence in other cards, but I don't much care since it's in a keyword.

modargo:
> I guess that'll solve #731 then. Personally, I don't think it was needed since we had precedence in other cards, but I don't much care since it's in a keyword.

I don't have a strong opinion on this but I figured it was mildly helpful at best and almost entirely harmless at worst.

I made the change before knowing that the Github issue existed (I never even thought to check for one), so I wouldn't mind if you decided to revert it or something. Either way it's a very minor thing.

PR #746, opened 2025-05-18
Energized pack: Make Investor account for Amplify costs

ActuallyEnbeon:
This PR also adds a new method to Wiz so effects similar to Investor are easy to make in future


modargo (approved):
Looks good, thanks for fixing this.

PR #747, opened 2025-05-26
Tweaked some cards in the Rimworld pack

Laugic:
Reworked Burning Passion: 2 Cost Rare Skill. Exhaust [your hand](any number of cards). Gain E and draw a card for each card Exhausted. Exhaust.
Reworked Tainted Weapon: 0 Cost Common Attack. Deal 8(11) damage. Put this card in the enemy if they are not dead.
Reworked Eat Without Table: 0 Cost Common Skill. Gain 4(6) Block. Gain 3(4) Temp HP. Add a Despair to your hand.
Reworked Kill Thirst: 1 Cost Uncommon Attack. Deal 8(10) damage. If this card doesn't kill, double its damage and shuffle a Despair into your draw pile.


src/main/resources/anniv5Resources/localization/eng/rimworldpack/Cardstrings.json
```diff
@@ -38,14 +38,12 @@
...
+    "DESCRIPTION": "Exhaust your hand. NL Gain [E] \u00a0and draw 1 card for each card Exhausted. NL Exhaust.",
+    "UPGRADE_DESCRIPTION": "Exhaust any number of cards. Gain [E] \u00a0and draw 1 card for each card Exhausted. NL Exhaust."
   },
   "${ModID}:TaintedWeapon": {
     "NAME": "Tainted Weapon",
-    "DESCRIPTION": "Deal !D! damage. NL Startup: Put this card in a random monster.",
-    "EXTENDED_DESCRIPTION": [
-      "Deal !D! damage."
-    ]
+    "DESCRIPTION": "Deal !D! damage. NL Put this card in the enemy if they are not dead."
```
modargo, line 46: Might make sense to use similar "if this doesn't kill" wording to Kill Thirst above. Combined with the comment I made about Kill Thirst's wording, that would make Tainted Weapon "If this doesn't kill, put it in the enemy."

It's the same idea in both cases (an effect that happens if it doesn't kill), so nice to use the same wording.

src/main/resources/anniv5Resources/localization/eng/rimworldpack/Cardstrings.json
```diff
@@ -22,7 +22,7 @@
   },
   "${ModID}:KillThirst": {
     "NAME": "Kill Thirst",
-    "DESCRIPTION": "Deal !D! damage. NL If this doesn't kill the enemy, add a *Despair to your hand."
+    "DESCRIPTION": "Deal !D! damage. If this card doesn't kill, double its damage and shuffle a *Despair into your draw pile."
```
modargo, line 25: Going by Sunder, you can omit the word "card" from this. Sunder if "if this kills an enemy", so "if this doesn't kill" seems like the closest wording to the base game for this (allowing for "an enemy" to be left out to save space).

modargo (commented):
The new cards look good. There's some dead code to clean up and a few small changes to make. After that, should be good to merge. Excited to play with the changed cards, I think this will make the pack a fair bit more interesting!

src/main/java/thePackmaster/cards/rimworldpack/TaintedWeapon.java
```diff
@@ -2,70 +2,34 @@
...
     public TaintedWeapon() {
         super(ID, 0, CardType.ATTACK, CardRarity.COMMON, CardTarget.ENEMY);
-        damage = baseDamage = 10;
-        magicNumber = baseMagicNumber = 3;
+        damage = baseDamage = 8;
     }
 
     public void use(AbstractPlayer p, AbstractMonster m) {
         addToBot(new DamageAction(m, new DamageInfo(p, damage, DamageInfo.DamageType.NORMAL), AbstractGameAction.AttackEffect.SLASH_HEAVY));
+        if(!m.isDeadOrEscaped() && !m.isDying && m.currentHealth > 0)
```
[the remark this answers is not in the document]
Laugic: I wanted to just skip the action if some other interaction with other modded powers / cards caused the creature to die upon the card being played or something, but since the checks are in the action I can try removing them here.

src/main/resources/anniv5Resources/localization/eng/rimworldpack/Cardstrings.json
Laugic: The inclusion of "card" I felt was necessary to allow the use of "its" in "double its damage". "If this doesn't kill, double its damage" sounds like it could even be referring to the enemy's damage.

Laugic: otherwise I would need to say "If this doesn't kill, double this card's damage" which makes the phrase longer than originally

PR #748, opened 2025-06-04
Various packs: tweak wording of exhausting cards from your hand to match the base game (allowing for some cards to diverge slightly to let them use more natural wording)

modargo:


modargo:
> Not sure if this one was a good idea. Might mess up the formatting on cards with random exhaust because of the extra length

Text length is a good point. Looking at some of the cards changed by this, the base game wording often adds an extra line. True Grit (the direct base game precedent) can afford that, and so can the simpler cards where the random exhaust is the last line of text (e.g. Soulfire and Doomguard from the Warlock pack). But it looks pretty ugly for cards like Midnight Oil and Hollower where the random exhaust is at the start. Especially since those cards are already 4 or 5 lines.

I'll change the wording for all these back to using "random card" instead of "card at random". I'm happy as long as Packmaster is internally consistent. And diverging from the base game slightly with this is fine since (1) nobody is ever going to be confused by this wording and (2) it makes the difference between one line of text and two lines of text.

modargo:
> I'll change the wording for all these back to using "random card" instead of "card at random".

This has been done.

